PR #638, opened 2017-10-10
detele empty device list from query params

### LINKED ISSUE #625: Incorrect request when showing telemetry for All Devices
# Type of issue

- [x] Bug
- [ ] New feature
- [ ] Improvement

# Description

When opening the dashboard for the first time, the UI makes this call:

```
GET /telemetry/v1/messages?from=NOW-PT1M&to=NOW&order=desc&devices=d1,d2,d3....all devices...dN
```

First of all, the list of devices may be too long when showing all the devices, and anyway the "devices" filter parameter should not be used when asking for **all** devices, because there is no filtering to run.

The correct request for **all devices** is to not use the 'devices' filter:

```
GET /telemetry/v1/messages?from=NOW-PT1M&to=NOW&order=desc
```

This is also one reason why the dashboard doesn't show any data, when the the list of device groups is missing. In such case call made is the following (note the empty list of device ids):

```
GET /telemetry/v1/messages?from=NOW-PT1M&to=NOW&order=desc&devices=
```

which results in an empty responses because the filter is empty.

maple10001:
# Types of changes

# Checklist:

# Description of the change

- detele empty device list from query params

# Motivation for the change

- fix #625


src/common/apiService.js
```diff
@@ -26,6 +26,18 @@ function postGroupData(group) {
   return data;
 }
 
+function filterParamObject(params) {
```
spryor: Add a JSDoc comment on this method explaining what it is for.

PR #643, opened 2017-10-10
update initail API calls for all devices

### LINKED ISSUE #625: Incorrect request when showing telemetry for All Devices

maple10001:
# Types of changes

# Checklist:

# Description of the change

- update initail API calls for all devices

# Motivation for the change

- fix #625


src/actions/deviceActions.js
```diff
@@ -81,24 +96,40 @@ export const loadDevicesByTelemetryMessages = () => {
...
+    return deviceList === lang.ALLDEVICES
+      ? ApiService.getAlarms({
+          from: `NOW-${timeRange}`,
+          to: 'NOW',
+          order: 'desc'
+        })
+          .then(data => {
+            dispatch({
+              type: types.LOAD_DEVICE_TELEMETRY_SUCCESS,
+              data: data
```
spryor: just data

src/actions/deviceActions.js
```diff
@@ -81,24 +96,40 @@ export const loadDevicesByTelemetryMessages = () => {
...
+      : ApiService.getAlarms({
+          from: `NOW-${timeRange}`,
+          to: 'NOW',
+          devices: deviceList,
+          order: 'desc'
+        })
+          .then(data => {
+            dispatch({
+              type: types.LOAD_DEVICE_TELEMETRY_SUCCESS,
+              data: data
```
spryor: just data

PR #657, opened 2017-10-13
field and operator fix

swetha571:
# Types of changes

# Checklist:

- [ ] The code follows the code style and conventions of this project.

# Description of the change
Field and operator fix in rules and actions flyout.
...

# Motivation for the change

...


src/components/ruleEditor/ruleEditor.js
```diff
@@ -168,14 +168,17 @@ class RuleEditor extends React.Component {
...
-                    if (!fields.some((o) => o.value === extract)) {
-                        fields.push({
-                            label: extract,
-                            value: extract
-                        });
-                    }
+                Object.values(telemetry).forEach(field => {
+                    const extract = field.MessageSchema.Fields;
+                    Object.keys(extract).forEach(field => {
+                        if (field.indexOf('_') !== -1) return; //we don't want keys that contain _
```
timlaverty: Why are we using the underscore convention to filter?  Are we sure variables we want in here will never have underscore?  This pushes a convention @ minimum to the Telemetry service, but it probably bleeds to telemetry-agent and simulation.

spryor: pasted image at 2017_10_13 11_31 am

The underscore is being used to help filter the Fields object. As I understand it, we don't need the `_unit` keys. It appears that there is some convention already being used in the backend. Are there issues with this assumption?

swetha571: Yes. Devis also confirmed that, which ever fileds key is ends with '_unit' or value is text will be eliminated from the field dropdown.

PR #669, opened 2017-10-16
fix bug for editing device Tags

maple10001:
# Types of changes

# Checklist:

# Description of the change

- removing addtional properties introduced when edting existing Tags

# Motivation for the change

- bug fix


src/components/flyout/deviceTagFlyout.js
```diff
@@ -215,10 +215,7 @@ class DeviceTagFlyout extends React.Component {
     const { devices } = this.props;
     const { newTags, deletedTagNames } = this.state;
     const deviceIds = devices.map(({ Id }) => `'${Id}'`).join(',');
-    const tags = {
-      ...this.state.commonTagValues,
-      ...this.state.overiddenDeviceTagValues
-    };
+    const tags = { ...this.state.commonTagValues };
```
spryor: I think this code could be made a bit cleaner. Change the `deleteExistingTag` method to store the results in an object where the key is the tag and the value is null. If you do that, lines 218 through 228 become the following (note adding commonTagValues to line 216).

```
const { newTags, deletedTagNames, commonTagValues } = this.state;
...
const tags = { 
   ...commonTagValues,
   ...deletedTagNames,
   ...newTags
};
```

src/components/flyout/deviceTagFlyout.js
```diff
@@ -305,7 +299,7 @@ class DeviceTagFlyout extends React.Component {
     const { commonTags, commonTagValues, commonTagTypes, deletedTagNames } = this.state;
     return (
       <div className="common-tags">
-        {commonTags.filter(tagName => deletedTagNames.indexOf(tagName) === -1).map((tagName, idx) => {
+        {commonTags.filter(tagName => deletedTagNames[tagName] !== null).map((tagName, idx) => {
```
spryor: `tagName in deletedTagNames`

src/components/flyout/deviceReconfigureFlyout.js
```diff
@@ -70,7 +70,7 @@ class DeviceReconfigureFlyout extends React.Component {
           .fromPromise(ApiService.getJobStatus(jobId))
           // Get completed jobs
           .filter(({ status }) => status === 3)
-          .flatMap(_ => deviceIds)
+          .flatMap(_ => deviceIds.filter(deviceId => devices.some(({ Id }) => deviceId === Id)))
```
spryor: 1) I think you should process the list of devices when it updates to create a helper data structure like a set of ids. Check for membership in the set instead of searching, potentially, the entire devices list multiple times.

2) Don't do the filter inside the flat map. Do it as part of the observable stream.
```
.flatMap(({ jobId, deviceIds }) =>
   ...
   .flatMap(_ => deviceIds)
)
.distinct()
.filter(deviceId => deviceIds.has(deviceId))
```

src/components/flyout/deviceTagFlyout.js
```diff
@@ -145,21 +145,24 @@ class DeviceTagFlyout extends React.Component {
...
       .flatMap(({ jobId, deviceIds }) =>
         Rx.Observable
           .fromPromise(ApiService.getJobStatus(jobId))
           // Get completed jobs
           .filter(({ status }) => status === 3)
-          .map(({ updateTwin }) => ({
-            tags: updateTwin.tags,
-            deviceIds
-          }))
+          .flatMap(_ => deviceIds.filter(deviceId => devices.some(({ Id }) => deviceId === Id)))
```
spryor: Same comment as above here.

PR #673, opened 2017-10-16
update device properties after job is done

### LINKED ISSUE #594: CRUD for Properties doesn’t appear to work
# Type of issue

- [ ] Bug
- [ ] New feature
- [ ] Improvement

# Description
Two problems:

When we add or change properties and click the apply button there's no indicator that the properties  are being saved. Apply should function as the other dialogs do (spinner) .
The redux cache entry for the device isn't invalidated. When the flyout is shown hubmanager should be requeried for just the device(s) edited if the cache is invalid. To see the updated data right now you need to refresh the page.

maple10001:
# Types of changes

- [x] New feature (non-breaking change which adds functionality)

# Checklist:

# Description of the change

- update device properties after job is done

# Motivation for the change

- fix #594


src/components/flyout/deviceReconfigureFlyout.js
```diff
@@ -32,14 +42,46 @@ class DeviceReconfigureFlyout extends React.Component {
...
+    const { devices, propertyUpdateJobs } = nextProps;
+    if (!_.isEqual(propertyUpdateJobs, this.props.propertyUpdateJobs)) {
+      this.checkJobStatus(devices, propertyUpdateJobs);
+    }
+    this.calcCommonConfiguration(devices);
+  }
+
+  checkJobStatus (devices, propertyUpdateJobs) {
+    const jobs = getRelatedJobs(devices, propertyUpdateJobs);
+    const jobStream = Rx.Observable.from(jobs);
```
spryor: No real need to save this as its own variable. I would just use the observable directly.

src/components/flyout/deviceReconfigureFlyout.js
```diff
@@ -32,14 +42,46 @@ class DeviceReconfigureFlyout extends React.Component {
...
+    jobStream
+      .flatMap(({ jobId, deviceIds }) =>
+        Rx.Observable
+          .fromPromise(ApiService.getJobStatus(jobId))
+          // Get completed jobs
+          .filter(({ status }) => status === 3)
+          .map(job => ({...job, deviceIds}))
```
spryor: Doesn't look like you care about the job at this point so `.flat(_ => deviceIds)` should be enough. (if you flatMap an array it will have the same effect as doing a flatMap of Rx.Observable.from(array)). I would also break out of the first flatMap at this point.

src/components/flyout/deviceReconfigureFlyout.js
```diff
@@ -32,14 +42,46 @@ class DeviceReconfigureFlyout extends React.Component {
...
+          // Get device from iothubmanager
+          .flatMap(({ deviceIds }) => Rx.Observable
+            .from(deviceIds)
+            .flatMap(deviceId => Rx.Observable
+              .fromPromise(ApiService.getDeviceById(deviceId))
+            )
+            .reduce((acc, cur) => acc.concat([cur]), [])
+          )
+      )
+      .subscribe(
```
spryor: I think this could be a bit cleaner. Is there any chance that device Ids could be duplicated in this stream?

```
Rx.Observable.from(jobs)
  .flatMap(({ jobId, deviceIds }) =>
    Rx.Observable
      .fromPromise(ApiService.getJobStatus(jobId))
      .filter(({ status }) => status === 3)
      .flatMap(_ => deviceIds)
  )
  .flatMap(deviceId => 
    Rx.Observable
      .fromPromise(ApiService.getDeviceById(deviceId))
  )
  .reduce((devices, device) => [...devices, device], [])
  .subscribe(
    devices => this.props.actions.updateDevices(devices),
    error => console.log('error', error)
  );
```

src/components/flyout/deviceReconfigureFlyout.js
```diff
@@ -32,14 +42,43 @@ class DeviceReconfigureFlyout extends React.Component {
...
+    if(!devices || !propertyUpdateJobs || !devices.length || !propertyUpdateJobs.length) return;
+    const jobs = getRelatedJobs(devices, propertyUpdateJobs);
+    Rx.Observable.from(jobs)
+      .flatMap(({ jobId, deviceIds }) =>
+        Rx.Observable
+          .fromPromise(ApiService.getJobStatus(jobId))
+          // Get completed jobs
+          .filter(({ status }) => status === 3)
+          .flatMap(_ => deviceIds)
+      )
```
spryor: .distinct()

PR #705, opened 2017-10-23
Enable start/stop simulations

spryor:
# Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)

# Description of the change
Add settings flyout
Add enable/disable simulations to the flyout
Remove Observable.fromPromise from locations where it isn't needed


src/components/settingFlyout/settingFlyout.scss
```diff
@@ -0,0 +1,38 @@
...
+
+  }
+
+  .btn-container {
+    display: flex;
+    justify-content: flex-end;
+    height: 30px;
+    margin-top: 48px;
+  }
+}
```
maple10001: nit: new line

Issue #584, opened 2017-10-02
Deleting devices


timlaverty:
# Type of issue

- [ ] Bug
- [ ] New feature
- [ ] Improvement

# Description
Attempting to delete a device shows:

![image](https://user-images.githubusercontent.com/5084013/31098905-12bf3fb2-a779-11e7-9c47-cc8b20531790.png)

CaitlinEsworthy:
Please see the attached mockups for the Delete device flow
![artboard 4](https://user-images.githubusercontent.com/30060166/31294466-d5c98bac-aa8f-11e7-9478-ed0e209bff29.png)
![artboard 5](https://user-images.githubusercontent.com/30060166/31294467-d5cdb81c-aa8f-11e7-85e9-1975aad2122b.png)
![artboard 6](https://user-images.githubusercontent.com/30060166/31294468-d5ea34e2-aa8f-11e7-8b38-d5ad609c68e8.png)

timlaverty:
@CaitlinEsworthy @hegate what do we do for alarm, method job, and property job detail pages that touch deleted devices?  Should the corresponding alarms get deleted?  For jobs we will have to look to see if they can be deleted from the hub.

CaitlinEsworthy:
@timlaverty When you say "corresponding alarms" do you mean alarms that are applied only to the devices deleted? I think the rule (if applied to multiple devices) then the rule should continue to exist but only on the non-deleted devices. But that begs the question: should we mention in the summary the rules effected or anything that is currently running (method)?

timlaverty:
Let's take deletion to an in person discussion.  I don't think there's significant complexity but would like to be sure we have our bases covered.  Hector, can you drive?

timlaverty:
For MVP:
1.	We will not allow deletion of individual simulated devices, we will allow deletion of physical devices.  
2.	The Delete button will show, “Coming Soon” when a simulated device is checked.
3.	The Delete button will be enabled when one or more physical devices are checked.  We will use the hub manager directly to delete those devices.
4.	The Delete button will show, “Coming Soon” when simulated and physical devices are checked.

PR #724, opened 2017-10-25
fix delete device flyout description

### LINKED ISSUE #719: Text on Delete Device Dialog is duplicated
# Type of issue 

- [ x] Bug
- [ ] New feature
- [ ] Enhancement

# Description 
Make sure text on the delete dialog matches the design when a simulated and physical device are both selected.  Right now two strings are duplicated:
'''	Deleting simulated devices is not currently supported, this action will only be taken on the physical devices.'''
below it:
'''	Deleting simulated devices is not currently supported, this action will only be taken on the physical devices.'''

maple10001:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

fix #719 

...

**Checklist:**

- [ ] All tests passed


src/components/flyout/deviceDeletionFlyout.js
```diff
@@ -61,7 +61,7 @@ class DeviceDeletionFlyout extends React.Component {
             {lang.DELETE_DEVICE}
           </div>
           <div className="content-description">
-            {IsSimulated ? lang.NOT_ABLE_TO_DELETE : lang.DELETE_DEVICE_DESCRIPTION}
+            {lang.DELETE_DEVICE_DESCRIPTION}
```
spryor: There should always be text on the delete page. The notification box should only appear when simulated devices are being ignored. See the design from https://github.com/Azure/pcs-remote-monitoring-webui/issues/584

PR #737, opened 2017-10-26
Feature/show version in settings

spryor:
# Type of change? 

- [x] Bug fix
- [x] New feature
- [ ] Enhancement

Show a (temporarily) hardcoded version number in the settings.
Fix issue in the error modal that wouldn't handle error objects correctly.


src/components/settingFlyout/settingFlyout.js
```diff
@@ -87,6 +87,9 @@ class SettingFlyout extends React.Component {
     return (
       <div className="setting-workflow-container">
         <div className="setting-section">
+          <div className="section-header">{ 'Version 0.0.1' }</div>
```
dluc: 0.0.1?

spryor: Placeholder value, cleared with Hector.

timlaverty: I would like to use the tag from the top level repos as the version number.  For Monday this will be `1.0.0-preview.3`.  Can you change it to this?

PR #739, opened 2017-10-26
After login redirect to /

dluc:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Fix deep linking issue

**Checklist:**

- [ ] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/739)


src/common/auth.js
```diff
@@ -87,6 +87,24 @@ function getAccessToken(callback) {
   );
 }
 
+/**
+ * Calculate the URI where to redirect the user after logging in/out.
+ * Note: won't work if the site is hosted in a subfolder.
+ */
+function getRedirectUri() {
```
spryor: Wouldn't `window.location.origin` do the same thing as this method?

dluc: that makes sense, wasn't aware of that function

PR #741, opened 2017-10-27
Change version label

spryor:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

Change the version label in the settings flyout


src/components/settingFlyout/settingFlyout.js
```diff
@@ -23,6 +23,8 @@ const currSimulationLabel = {
   false: lang.STOPPED
 };
 
+const version = '1.0.0-preview.3';
```
timlaverty: I think in the future it'd be better to read this from a file on disk.  That & we should think about whether we list the container tags in it instead of the top level repo tag.  For not it looks fine.

dluc: could we have this value in `config.js` ?

dluc: we have config.js that I think we could reuse

spryor: Yep, its on the todo. The config would be an easy place to put it.

spryor: Done.

PR #749, opened 2017-10-27
Eliminating the spaces and special chars for JobInput

swetha571:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Eliminating spaces and special chars if even user imputed.
Added input text sample.

...

**Checklist:**

- [ ] All tests passed


src/common/lang.js
```diff
@@ -135,6 +135,7 @@ const lang = {
   INFO: 'Info',
   JOBNAME: 'Job Name',
   JOB_NAME: 'Job Name',
+  JOB_NAME_REFERENCE: 'Job Name will not display spaces and special characters',
```
spryor: I think this string makes more sense in the context of a form. 

`JOB_NAME_REFERENCE: 'Only letters and numbers are allowed in the job name',`

src/components/flyout/deviceReconfigureFlyout.js
```diff
@@ -190,8 +190,10 @@ class DeviceReconfigureFlyout extends React.Component {
     this.state.commonConfiguration.forEach(item => {
       reportedProps[item.label] = item.value;
     });
+    let jobName = this.state.jobInputValue ;
```
spryor: `const jobName = this.state.jobInputValue.replace(/\W/g, "");`

Note the change in the regex. You should also pull this regex into a utility file that is imported across all the other places it is used.

src/components/flyout/deviceReconfigureFlyout.js
```diff
@@ -190,8 +190,10 @@ class DeviceReconfigureFlyout extends React.Component {
...
+    jobName = jobName.replace(/[\W_]/g, "");
```
spryor: Instead of doing the replace in the `apply`, you should do in the onChange for the input fields. That way the user can see in realtime what their job name will be. It's a better user experience with less room for confusion.

src/common/lang.js
swetha571: These are the comps provided by Caitlin.
image

src/components/flyout/deviceReconfigureFlyout.js
swetha571: I think, (/[\W_]/g, "") this would be the right regex code, because we have to eliminate the _ and  [] also.

PR #759, opened 2017-10-30
ruleName added to deviceDetailFlyout

swetha571:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Rulename added to Device Detail Flyout.

...

**Checklist:**

- [ ] All tests passed


src/components/flyout/deviceDetailFlyout.js
```diff
@@ -121,11 +124,16 @@ class DeviceDetailFlyout extends Component {
...
+  componentWillMount(){
+    this.props.actions.loadRulesList();
+  }
+
   componentWillUnmount() {
     this.subscriptions.forEach(sub => sub.unsubscribe());
   }
 
   componentWillReceiveProps(nextProps) {
+    this.props.actions.loadRulesList();
```
spryor: Do we need to call this on every new props? This will be called a lot as is.

src/components/flyout/deviceDetailFlyout.js
```diff
@@ -176,7 +184,13 @@ class DeviceDetailFlyout extends Component {
       this.setState({
         showAlarmsGrid: data.Items.length > 0,
         alarmRowData: data.Items.map(item => {
+          this.props.rulesAndActions.forEach(Rule => {
```
spryor: As mentioned in the PR for the alarm list, you should preprocess the rulesAndActions when new props come in. It should create a map of `id` to `name` to avoid looping the rulesAndActions for every alarm.

PR #761, opened 2017-10-30
Ss.bing map issue for standard deployment

swetha571:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Standard deployment Bingmap keys. Even if the BingMapKey is empty sting, we should still show the StaticBingMap.

...

**Checklist:**

- [ ] All tests passed


src/components/deviceMap/deviceMap.js
```diff
@@ -167,6 +167,18 @@ class DeviceMap extends Component {
     });
   }
 
+  renderStaticMap() {
+      return <div className="map-container">
+        <div className="static-map-modal">
+          <span className="static-text">
+            {lang.MAP_MODAL_CONTENT}
+            <a className="static-link" rel="noopener noreferrer" href="https://docs.microsoft.com/en-us/azure/iot-suite/iot-suite-faq" target='_blank'>{lang.MAP_MODAL_FAQS}</a>
```
dluc: please remove the "en-us" from the URL

src/components/deviceMap/deviceMap.js
```diff
@@ -181,15 +193,8 @@ class DeviceMap extends Component {
...
-            ? <div className="map-container">
-                <div className="static-map-modal">
-                  <span className="static-text">
-                    {lang.MAP_MODAL_CONTENT}
-                    <a className="static-link" rel="noopener noreferrer" href="https://docs.microsoft.com/en-us/azure/iot-suite/iot-suite-faq" target='_blank'>{lang.MAP_MODAL_FAQS}</a>
-                  </span>
-                </div>
-                <img src={StaticMapWithOutDevices} alt="StaticMap" className="static-bing-map"/></div>
-            : <div id="deviceMap" className="dashboard_device_map" />) : null }
+            ? this.renderStaticMap()
```
dluc: could you simplify the condition? something like
```
BingMapKey && BingMapKey !== config.STATUS_CODES.STATIC ? ...bingmap... : ...static map...
```

src/components/deviceMap/deviceMap.js
swetha571: Just for making sure. In comps they have given me this " https://docs.microsoft.com/en-us/azure/iot-suite/iot-suite-faq". Even if i eliminate the "en-us" it s still redirecting to same link with en-us appending. Is that the actula behaviour??

swetha571: I am on it.

PR #762, opened 2017-10-31
AllFlyoutsBrokenOnAlarmsDetailsPage

swetha571:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Fixed all flyouts in alarm Details Page

...

**Checklist:**

- [ ] All tests passed


src/components/maintenance/ruleDetails.js
```diff
@@ -186,12 +186,7 @@ class RuleDetailsPage extends Component {
       //TODO: finish all the event handlers
       rowData: this.state.devicesGridData,
       domLayout: 'autoHeight',
-      softSelectId: this.state.softSelectedDeviceId,
-      getSoftSelectId: btnActions.getSoftSelectId,
```
spryor: If you remove this, the soft selection state will not work. Why does this need to be removed? Are you sure you need to remove all of this?

swetha571: Yes. I am very sure. From morning investigated various ways. Later found the event handlers are causing the issue.

timlaverty: Does it break soft selection?  i.e. we can't have hard select make the schedule/tags/properties dialogs work and in the process break the device details flyout.

spryor: It doesn't break soft selection as the flyout still on opens. It appears to break the soft select effect on the row (the highlight). As it turns out, the soft select effect on this page is already broken since it wasn't implemented properly.

swetha571:
I fixed the softSelect. I removed this because "onHardSelectChange: btnActions.onDeviceGridHardSelectChange," this event is coming from devicesGrid basic  (
image)functionality.

PR #785, opened 2017-11-01
Attempt to resolve telemetry memory leak

spryor:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

The telemetry chart appears to be causing a memory leak (presumably caused by D3 within the C3 library). This fix places a timer on the chart to rebuild it every 5 minutes to clean up any references to old data avoiding garbage collection


src/components/charts/timeline.js
```diff
@@ -4,39 +4,75 @@ import React, { Component } from 'react';
...
-    this.state = {
-      timeline: {}
-    };
+    this.chart = {};
+    this.chartDataQueue = [];
   }
 
   componentDidMount() {
-    this.setState({ timeline: C3.generate({ ...this.props.chartConfig }) });
+    // Every five minutes, rebuild the chart to avoid a memory leak in the chart library
```
timlaverty: Can you add a GitHub Issue that maps to this - we should explain that this is a workaround b/c of a bug in D3.

PR #849, opened 2017-12-08
Enable/disable rule detail flyout

swetha571:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Enable/Disable rule details flyout

...

**Checklist:**

- [ ] All tests passed


src/components/ruleOverview/ruleOverview.js
```diff
@@ -64,10 +64,8 @@ class RuleOverviewFlyout extends Component {
                 <div className="clearfix"/>
                 <div className="rule-overview-header">
                     <div className="toggle-button">
-                        <div onClick={this.onToggleStatus}
-                             className={`icon ${!this.state.enabled ? 'icon-enable' : 'icon-disable'}`}/>
-                        <div
-                            className="text">{this.state.enabled ? lang.ENABLESELECTEDRULE : lang.DISABLESELECTEDRULE}</div>
+                        <div className={`icon ${this.state.enabled ? 'icon-disable': 'icon-enable'}`}/>
+                        <div className="text">{this.state.enabled ?  lang.DISABLESELECTEDRULE : lang.ENABLESELECTEDRULE}</div>
```
spryor: There is an extra space after the `?`

src/components/ruleOverview/ruleOverview.js
```diff
@@ -17,15 +17,15 @@ class RuleOverviewFlyout extends Component {
         };
```
dluc: is this equivalent to:
```
this.state = {
  enabled: props.rules.some(rule => !rule.Enabled)
}
```

src/components/ruleOverview/ruleOverview.js
```diff
@@ -17,15 +17,15 @@ class RuleOverviewFlyout extends Component {
...
-        this.setState({enabled: !this.state.enabled});
-    };
-
     onSave = () => {
         let promises = [];
         let updatedRules = [];
         this.props.rules.forEach(rule => {
-            if (rule.Enabled !== this.state.enabled) {
+            if (rule.Enabled === this.state.enabled) {
+                rule.Enabled = !this.state.enabled;
```
dluc: is this equivalent to `rule.Enabled = false` ? if so I'd change it

src/components/ruleOverview/ruleOverview.js
swetha571: No. If the rule is already disabled - then when we open the flyout, we see the enable toggle.
If the rule is already ENABLED -  then when we open the flyout , we see the disable toggle with out user toggles them.

src/components/ruleOverview/ruleOverview.js
```diff
@@ -17,15 +17,15 @@ class RuleOverviewFlyout extends Component {
...
+                updatedRules.push(rule);
```
spryor: The rule is pushed into the array and the promise is pushed into its array regardless of the if condition. Move these values out of the if and do them all the time.

PR #851, opened 2017-12-20
hide the pagination if grid rows are less

swetha571:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

- Hide the pagination when Number of rows is less as per the requirement. Devices, rules page grids should have 50 rows per page, if the rows are less than 50 per page then we don't show the pagination.
- AlarmGrid should have 5 rows per systemalarm panel. If they exceed more than 5 then we show pagination.

...

**Checklist:**

- [ ] All tests passed


src/components/alarmList/alarmList.js
```diff
@@ -133,7 +133,8 @@ class AlarmList extends Component {
         error={this.state.error}
         title={lang.ALARMSTATUS}>
         <div className="grid-container">
-          <AlarmsGrid {...alarmsGridProps} />
+          {this.state.rowData &&
+          <AlarmsGrid {...alarmsGridProps} pagination = {this.state.rowData.length > 8 ? true: false}/>}
```
dluc: is "8" the right value in all the situations? if I remember correctly the number of alarms in a page depends on the browser window size, the zoom settings etc.

src/components/alarmList/alarmsGridConfig.js
```diff
@@ -8,7 +8,7 @@ import { gridValueFormatters } from '../pcsGrid/pcsGridConfig';
 
 const { checkForEmpty } = gridValueFormatters;
 
-export const DEFAULT_ALARM_GRID_PAGE_SIZE = 8;
+export const DEFAULT_ALARM_GRID_PAGE_SIZE = 5;
```
dluc: in `alarmList.js` the value is 8, here it's 5. Are they supposed to be different or the same?   Also, how did you calculate the value before (8) and now (5) ?

src/components/pages/devices/devices.js
```diff
@@ -86,7 +86,8 @@ class DevicesPage extends Component {
             </div>
             <div onClick={this.refreshData} className="refresh-icon icon-sm" />
           </div>
-          <DevicesGrid {...deviceGridProps} />
+          {devices &&
+          <DevicesGrid {...deviceGridProps} pagination={(devices|| []).length > 50 ? true :  false}/>}
```
dluc: what is "50" in this context and how was the value calculated? could you use a constant instead of hardcoding the value?

src/components/rulesActionsList/ruleAndActionsConfig.js
```diff
@@ -9,6 +9,8 @@ import lang from "../../common/lang";
 
 export const LastTriggerDefaultWidth = 310;
 
+export const DEFAULT_RULES_GRID_PAGE_SIZE = 50;
```
dluc: why is this constant all upper case, while the previous one is PascalCased? we should fix the inconsistencies.  Also please add a comment describing how the value (50 in this case) is calculated, otherwise it will be pretty much impossible to maintain it

src/components/rulesActionsList/rulesActionsList.js
```diff
@@ -54,9 +54,11 @@ class RulesActionsList extends Component {
       ...this.props,
       onGridReady: this.onGridReady
     };
+    
     return (
       <div className="rules-actions-container">
-        <PcsGrid {...gridProps} />
+        {(this.props.rowData || []).length > 0 &&
+        <PcsGrid {...gridProps} pagination= {(this.props.rowData || []).length > 50 ? true : false}/>}
```
dluc: ditto

src/components/pages/devices/devices.js
swetha571: This value is given by Caitlin. Sure i will use the const instead of hard coding.

swetha571: This value is given by Caitlin. Sure instead of hard coded value i will add const.

src/components/rulesActionsList/ruleAndActionsConfig.js
swetha571: Sure. I will fix the casing issue but the 50 rows per page value is given by Caitlin.

src/components/alarmList/alarmsGridConfig.js
swetha571: It's 5. Because the rows goes beyond the alarmGrid panel if it exceeds 7.

swetha571:
Updated the PR with
- Removed hard coded value instead used const values.
- Fixed the inconsistencies const casing issues.
- Caitlin provided the Number of rows per page. I will find out from her about that.

src/components/rulesActionsList/rulesActionsList.js
```diff
@@ -54,9 +55,11 @@ class RulesActionsList extends Component {
       ...this.props,
       onGridReady: this.onGridReady
     };
+
     return (
       <div className="rules-actions-container">
-        <PcsGrid {...gridProps} />
+        {(this.props.rowData || []).length > 0 &&
```
spryor:
```
{
   ... multi line code
}
```

src/components/rulesActionsList/rulesActionsList.js
```diff
@@ -54,9 +55,11 @@ class RulesActionsList extends Component {
...
+        <PcsGrid {...gridProps} pagination= {(this.props.rowData || []).length > Config.DEVICES_RULESGRID_ROWS ? true : false}/>}
```
spryor: There is a space after `pagination=`

src/components/rulesActionsList/ruleAndActionsConfig.js
dluc: could you ask Caitlin how the constants were calculated?

also, this "50" is it the same of Config.DEVICES_RULESGRID_ROW, and could we reuse the Config value?

PR #856, opened 2018-01-08
Ss.maitenance action bar fix

swetha571:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Maintenance action bar fixed. when we click on the devices grid in maintenance details page and go back to the maintenance main page should reset the action bar. 
Disable the Device filter in Job details page.

...

**Checklist:**

- [ ] All tests passed


src/components/pages/maintenance/maintenance.js
```diff
@@ -335,7 +335,7 @@ class MaintenancePage extends Component {
     let breadcrumbs;
     const pathName = this.props.location.pathname;
     const alarmsGridData = this.props.alarmsGridData;
-    const parentLink = <Link to={'/maintenance'}>{lang.MAINTENANCE}</Link>;
+    const parentLink = <Link to={'/maintenance'} onClick={() => this.setState({ contextBtns: true })}>{lang.MAINTENANCE}</Link>;
```
spryor: Avoid creating static functions in the render method. That will create a new function instance each time render is called, causing the children to re-render when nothing actually changed.

swetha571: Sure.

src/components/pages/maintenance/maintenance.js
```diff
@@ -252,7 +252,10 @@ class MaintenancePage extends Component {
   }
 
   componentWillReceiveProps(nextProps) {
-    const { devices } = nextProps;
+    const { devices, params } = nextProps;
+    if (!params.id && !params.JobId) {
+      setTimeout(()=> this.setState({ contextBtns: true }));
```
spryor: Why this setTimeout?

PR #859, opened 2018-01-10
Expansion Chevron changed for device details flyout

swetha571:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

DeviceDetailFlyout Chevron icon changed based on the toggle feature. 

...

**Checklist:**

- [ ] All tests passed


src/components/flyout/drawer.scss
```diff
@@ -1,40 +1,48 @@
...
-      color: #ffffff;
-      img {
-        opacity: 1;
-      }
-    }
+      color: $white;
 
-    img {
-      opacity: 0.7;
+      img { opacity: 1; }
```
spryor: If the opacity is 100%, why do we need opacity at all?

src/components/flyout/drawer.js
```diff
@@ -30,7 +30,9 @@ class Drawer extends Component {
       <div className="drawer-wrapper">
         <div className="drawer-header" onClick={this.toggleHeader}>
           {this.props.title}
-          <img src={ChevronIcon} className="chevron-open" alt="ChevronIcon" />
+          {this.state.showContent
```
spryor: As a pattern, place multi line code blocks on seperate lines.
```
// Single line
{ ... some code }

// Multi line
{
    ... code
    ... more code
}
```

src/components/flyout/drawer.js
```diff
@@ -30,7 +30,11 @@ class Drawer extends Component {
       <div className="drawer-wrapper">
         <div className="drawer-header" onClick={this.toggleHeader}>
           {this.props.title}
-          <img src={ChevronIcon} className="chevron-open" alt="ChevronIcon" />
+          {
+            this.state.showContent
+            ? <img src={ChevronIcon} className="chevron-close" alt="ChevronIcon" />
```
spryor: "ChevronIcon" also isn't helpful. This string needs to be something that someone who is in some way impaired can make sense of from a functional point of view. Imagine that someone who is blind is navigating the site using a screen reader. When the reader reads "ChevronIcon", what does that tell them?

PR #875, opened 2018-02-21
Add simulation base arch

spryor:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

Update base arch
- Add CSS theming and mixins
- Restructure folder structure
- Update redux and redux-observable logic
- Use react router v4
- Add shared controls


src/components/app/main/main.js
```diff
@@ -2,16 +2,18 @@
...
-    return (
-        <div className="page-container">
-            {props.children}
-        </div>
-    );
-}
+const Main = (props) => (
+  <main className="app-main">
+    <div className="main-container">
+      { props.children }
```
ppathan: Is props.children a list or object? The reason i ask is because if it is just single object then props.child might be better name

src/components/app/navigation/navigation.js
```diff
@@ -0,0 +1,77 @@
...
+  toggleExpanded = (event) => {
+    this.setState({ collapsed: !this.state.collapsed });
+  }
+
+  render() {
+    const isExpanded = !this.state.collapsed;
+    return (
+      <nav className={`app-nav ${isExpanded && 'expanded'}`}>
+        <div className="nav-item company">
+          <NavIcon path={svgs.contoso} />
```
ppathan: Wondering if we should make this passed from props. So that we can easily update the logo from contoso to something else.

src/components/pages/dashboard/dashboard.container.js
```diff
@@ -0,0 +1,6 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+import { translate } from 'react-i18next';
+import { Dashboard } from './dashboard';
+
+export const DashboardContainer = translate()(Dashboard);
```
ppathan: Question: Is translate that gives us localization support?

src/components/pages/pageNotFound/pageNotFound.js
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+import React from 'react';
+
+import './pageNotFound.css';
+
+export const PageNotFound = () => (
```
ppathan: This doesn't need to be a component?

mechaffin:
The file structure feels inconsistent.  Sometimes the related files are next to each other (i.e. the component and its scss and test files are together in the same folder).  However, at other times, files are grouped by type instead (shared component js and scss are in different folders; also, services and their related models are in separate folders).  Pick one organizational scheme and stick with it throughout the application.

src/components/app/main/main.js
spryor: It's a react property. https://reactjs.org/docs/react-api.html#reactchildren

src/components/pages/dashboard/dashboard.container.js
spryor: Yes, it is a higher order component (like redux connect) that passes the translation methods through the props. https://react.i18next.com/components/translate-hoc.html

src/components/pages/pageNotFound/pageNotFound.js
spryor: It is. React components can be either class components or functional components. https://reactjs.org/docs/components-and-props.html#functional-and-class-components

src/services/httpClient.js
```diff
@@ -0,0 +1,133 @@
...
+ */
+export class HttpClient {
+
+  /**
+   * Constructs a GET ajax request
+   *
+   * @param {string} url The url path to the make the request to
+   */
+  static get(url, options = {}, withAuth = true) {
+    return HttpClient.ajax(url, { ...options, method: 'GET' }, withAuth);
```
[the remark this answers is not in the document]
spryor: As discussed, we can make them an enum but since these are never called explicitly I don't think it is worth it. The methods themselves essentially act as enums.

DigiBanks99:
All the old flyouts were removed during this commit. Might I ask why?

PR #876, opened 2018-02-28
Simple listing of device ids on the device page

mechaffin:
…the UI infrastructure: Service, Store, Reducer, Epic.

# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

Simple listing of device ids on the device page, utilizing layers of the new UI infrastructure: service, store, reducers, selectors, epics, etc.

**Checklist:**

- [X] All tests passed


public/locales/en/translations.json
```diff
@@ -9,5 +9,11 @@
     "devices": "devices",
     "rules": "rules",
     "maintenance": "maintenance"
+  },
+  "components": {
```
spryor: I would remove the components key, just use devices.

src/components/pages/devices/devices.container.js
```diff
@@ -1,6 +1,25 @@
...
+
+// Pass the devices status
+const mapStateToProps = state => ({
+    devices: getDevices(state),
+    error: getDevicesError(state)
+});
+
+// Wrap the dispatch method
+const mapDispatchToProps = dispatch => ({
+    refresh: () => dispatch(iotHubManagerEpics.actions.fetchDevices()),
```
spryor: These functions are the same only one should be passed. Maybe rename it to `fetchDevices`.

src/components/pages/devices/devices.js
```diff
@@ -1,14 +1,30 @@
...
 import React, { Component } from 'react';
+import { Btn } from 'components/shared';
 
 import './devices.css';
 
 export class Devices extends Component {
-  render () {
+  constructor(props) {
+    super(props);
+    if (props.initialize) {
```
spryor: Remove the if. If initialize isn't included then something is wrong and the component should die. Check the existence with [propTypes](https://reactjs.org/docs/typechecking-with-proptypes.html)

src/components/pages/devices/devices.js
```diff
@@ -1,14 +1,30 @@
...
+      this.props.initialize();
+    }
+  }
+
+  render() {
+    const { t, devices, refresh } = this.props;
+    var listItems = t('components.devices.noneFound');
```
spryor: Avoid using `var`. Default to using `const` and `let` when `const` is unsuitable. More about this below.

src/components/pages/devices/devices.js
```diff
@@ -1,14 +1,30 @@
...
+    if (devices != null && devices.length > 0) {
```
spryor: Use javascript truthyness to make this more succinct `if (devices && devices.length)`

src/components/pages/devices/devices.scss
```diff
@@ -8,3 +8,8 @@
   padding: 20px;
   @include rem-font-size(20px);
 }
+.device {
```
spryor: `.device` should be nested in `.devices-container`. Every component with css elements should be encapsulated in a container class to avoid polluting the global css namespace.

src/store/reducers/iotHubManagerReducer.js
```diff
@@ -0,0 +1,49 @@
+// Copyright (c) Microsoft. All rights reserved.
```
spryor: Name this file devicesReducer instead of referencing the service. It may be that the devices reducer ends up calling multiple services so there isn't a 1-to-1 mapping from reducers and services (or even reducers to pages).

src/store/reducers/iotHubManagerReducer.js
```diff
@@ -0,0 +1,49 @@
...
+
+import 'rxjs';
+import { Observable } from 'rxjs';
+import { IoTHubManagerService } from 'services';
+import { toDeviceListModel } from 'services/models';
+import { createReducerScenario, createEpicScenario } from 'store/utilities';
+
+// Simulation reducer constants
+const EMPTY_DEVICES = toDeviceListModel();
+const initialState = { data: undefined };
```
spryor: Rename `data` to `items`.

src/store/reducers/iotHubManagerReducer.js
```diff
@@ -0,0 +1,49 @@
...
+  fetchDevices: {
+    type: 'DEVICES_FETCH',
+    epic: () =>
+      IoTHubManagerService.getDeviceList()
+        .map(redux.actions.updateModel)
+        .startWith(redux.actions.clearModel())
+        .catch(devicesError)
+  }
+});
+// ========================= Epics - END
```
spryor: Needs a new line

src/store/reducers/iotHubManagerReducer.js
```diff
@@ -0,0 +1,49 @@
...
+// Simulation reducer constants
+const EMPTY_DEVICES = toDeviceListModel();
+const initialState = { data: undefined };
+
+// ========================= Reducers - START
+const deviceListModelReducer = (state, action) => ({ ...state, data: action.payload });
+const iotHubManagerErrorReducer = (state, action) => ({ error: action.payload });
+const initialStateReducer = (state, action) => initialState;
+
+const updateModel = { type: 'DEVICES_UPDATE', reducer: deviceListModelReducer };
```
spryor: The `model` terminology is a hold over from simulation (because we were saving a model). It doesn't make sense to use it here. `updateDevices` and `clearDevices` are clearer.

src/store/reducers/iotHubManagerReducer.js
```diff
@@ -0,0 +1,49 @@
...
+
+export const redux = createReducerScenario({
+  updateModel,
+  clearModel: { ...updateModel, type: 'DEVICES_CLEAR', staticPayload: EMPTY_DEVICES },
+  registerError: { type: 'DEVICES_ERROR', reducer: iotHubManagerErrorReducer },
```
spryor: I'm not going to change this for this PR but I want to note that I want to change how errors are handled from simulation. In simulation I would wipe the entire reducer state to an error state to keep things easy. That worked fine given how simple the simulation reducers are but I don't think it will scale here. We can talk about this later as I'm still doing some research into the best way to approach it. The same thing applies to the clear reducer. In simulation we avoided having a reducer for loading states by hiding those semantics in `undefined` values (e.g. an  `undefined` meant the data was loading). That works ok but we should probably be more explicit about loading states too.

src/components/pages/devices/devices.js
```diff
@@ -1,14 +1,30 @@
...
+      listItems = devices.map((device, index) =>
+        <li key={index} className="device">{device.id}</li>
+      );
+    }
     return (
       <div className="devices-container">
-        Devices Page
+        <ul>{listItems}</ul>
```
spryor: I wouldn't do it like this since listItems is now heterogeneous variable. That causes 1) a conflict in the naming (e.g. listItems might not be a list at all) and 2) could result in `ul` tags with just a string inside them, which is odd html. A better approach would be something like:

```
renderDevices(devices, noDevicesMessage) {
    if (devices && devices.length) {
        return render (
          <ul>
            { devices.map((device, index) => <li key={index} className="device">{device.id}</li>) }
          </ul>
        );
    }
    return noDevicesMessage
}
...
// in the render
{ this.renderDevices(devices, t('components.devices.noneFound'))}
```
Of some variation with the logic embedded in the jsx.

src/app.config.js
```diff
@@ -1,6 +1,10 @@
 // Copyright (c) Microsoft. All rights reserved.
 
 const Config = {
+  // TEMP: Base service urls
+  baseUrls: {
+    iotHubManager: "https://maryellen-0226.azurewebsites.net/iothubmanager/v1/"
```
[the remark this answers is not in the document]
mechaffin: How should we specify it for now?  Or jsut check in with an empty string?

src/services/iotHubManagerService.js
```diff
@@ -0,0 +1,18 @@
...
+
+const ENDPOINT = Config.baseUrls.iotHubManager;
+
+/** Contains methods for calling the Device service */
+export class IoTHubManagerService {
+
+  /** Returns a list of devices */
+  static getDeviceList() {
+    return HttpClient.get(`${ENDPOINT}devices`)
+      .do(response => console.log('response', response))
```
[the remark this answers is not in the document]
mechaffin: Thought I had caught all of those. Darn!

src/services/iotHubManagerService.js
```diff
@@ -0,0 +1,18 @@
...
+import { HttpClient } from './httpClient';
+import { toDeviceListModel, toDeviceModel } from './models';
+
+const ENDPOINT = Config.baseUrls.iotHubManager;
+
+/** Contains methods for calling the Device service */
+export class IoTHubManagerService {
+
+  /** Returns a list of devices */
+  static getDeviceList() {
```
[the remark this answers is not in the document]
mechaffin: Sure

src/services/iotHubManagerService.test.js
```diff
@@ -0,0 +1,27 @@
+// Copyright (c) Microsoft. All rights reserved.
```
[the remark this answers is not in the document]
mechaffin: OK, I was hoping to get it working... but that can also be another check in.

src/components/pages/devices/devices.js
```diff
@@ -15,7 +15,7 @@ export class Devices extends Component {
 
   render() {
     const { t, devices, refresh } = this.props;
-    var listItems = "No devices found.";
+    var listItems = t('components.devices.noneFound');
```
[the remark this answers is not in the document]
mechaffin: Sure

public/locales/en/translations.json
mechaffin: OK.

src/components/pages/devices/devices.js
mechaffin: I added it to make the test pass.  The component handles when the device list isn't passed, but blows up without the handler for the refresh button... so I made it more resilient to that.  
Alternatively, I can change the test.

src/components/pages/devices/devices.container.js
mechaffin: Haha.  You added the second one.  But yes, you're right... we shouldn't pass two of the same thing.

src/components/pages/devices/devices.js
mechaffin: OK

src/components/pages/devices/devices.scss
mechaffin: OK

src/store/reducers/iotHubManagerReducer.js
mechaffin: OK. That makes sense.  I'll learn how things interact better soon, I'm sure.  For now, I was going with the 1 to 1 mapping.

src/store/reducers/iotHubManagerReducer.js
mechaffin: Yeah, we've talked about how we might want error handling to be different.  This is really just a first pass... and there's not much here yet.  We'll figure out what we want overall... and then change to match that.

src/store/reducers/iotHubManagerReducer.js
mechaffin: OK

src/app.config.js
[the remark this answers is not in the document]
spryor: At this point we can just hard code the needed urls locally when we need them

src/components/pages/devices/devices.js
spryor: The fix there would be to fake the prop in the test.

src/app.config.js
```diff
@@ -1,6 +1,10 @@
 // Copyright (c) Microsoft. All rights reserved.
 
 const Config = {
+  // TEMP: Base service urls
+  baseUrls: {
+    iotHubManager: "https://YOURHOSTNAME/iothubmanager/v1/"
```
spryor: If you are going to keep the baseUrls here I could make the hostname a variable.

```
const baseURL = '';
...
const Config = {
    serviceEndpoints: {
        iotHubManager: `${baseURL}iothubmanager/v1/`
    }
    ...
}
```

src/components/pages/devices/devices.js
```diff
@@ -1,14 +1,33 @@
...
+
+  renderDevices(devices, noDevicesMessage) {
+    if (devices && devices.length) {
+      return (
+        <ul>
+          {devices.map((device, index) => <li key={index} className="device">{device.id}</li>)}
+        </ul>
+      );
+    }
+    return noDevicesMessage
```
spryor: missing `;`

src/app.config.js
mechaffin: done.

src/components/pages/devices/devices.js
mechaffin: done.

PR #879, opened 2018-03-06
Add rules page

spryor:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

Add the rulesGrid
Add rulesGrid cellRenderers
Add reusable refresh bar
Add calls to get rulesData


src/components/pages/rules/rulesGrid/rulesGrid.js
```diff
@@ -0,0 +1,54 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+import React, { Component } from 'react';
+import { translate } from 'react-i18next';
+import { PcsGrid } from 'components/shared';
+import { rulesColumnDefs, checkboxParams, defaultRulesGridProps } from './rulesGridConfig';
+import { isFunc } from 'utilities'; //translateColumnDefs
```
mechaffin: commented translateColumnDefs?

src/components/pages/rules/rules.js
```diff
@@ -1,14 +1,32 @@
...
+import RulesGrid from './rulesGrid/rulesGrid';
+import { Btn, RefreshBar } from 'components/shared';
 import './rules.css';
 
 export class Rules extends Component {
+  constructor(props) {
+    super(props);
+
+    if (!this.props.rules.length) {
+      this.props.fetchRules();
```
mechaffin: Need to call isFunc?

src/components/pages/rules/rules.js
```diff
@@ -1,14 +1,32 @@
...
+    }
+  }
+
+  changeDeviceGroup = () => {
+    const { changeDeviceGroup, deviceGroups }  = this.props;
+    changeDeviceGroup(deviceGroups[1].id);
```
mechaffin: isFunc?  Shouldn't we check that for everything we expect to be a function?

src/components/pages/devices/devices.js
```diff
@@ -2,23 +2,29 @@
...
 export class Devices extends Component {
   constructor(props) {
     super(props);
-    this.props.fetchDevices();
+    // this.props.fetchDevices();
+  }
+
+  changeDeviceGroup = () => {
+    const { changeDeviceGroup, deviceGroups }  = this.props;
+    changeDeviceGroup(deviceGroups[1].id);
```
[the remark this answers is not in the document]
spryor: Yep, just temporary for testing purposes. This will be replaced with an actual selection dropdown.

src/components/pages/devices/devices.js
```diff
@@ -2,23 +2,29 @@
...
-import { Btn } from 'components/shared';
+import { Btn, RefreshBar } from 'components/shared';
 
 import './devices.css';
 
 export class Devices extends Component {
   constructor(props) {
     super(props);
-    this.props.fetchDevices();
+    // this.props.fetchDevices();
```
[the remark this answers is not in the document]
spryor: Removed in next PR.

src/store/reducers/appReducer.js
```diff
@@ -1,18 +1,47 @@
...
     type: 'APP_INITIALIZE',
     epic: () => [
-      // redux.actions.updateDeviceModels(),
+      epics.actions.fetchDeviceGroups(),
+      redux.actions.updateActiveDeviceGroup()
     ]
   },
 
+  /** Get the account's device groups */
+  fetchDeviceGroups: {
```
mechaffin: It feels like this should be in deviceReducer.  The whole reason we called it "device" instead of "iotHubManager" was because we knew there would be mutliple services called to deal with devices.

src/store/reducers/devicesReducer.js
```diff
@@ -58,21 +73,22 @@ const fetchableTypes = [
...
 
 export const reducer = { devices: redux.getReducer(initialState) };
 // ========================= Reducers - END
 
 // ========================= Selectors - START
-export const accessReducer = state => state.devices;
-export const getEntities = state => accessReducer(state).entities;
-export const getItems = state => accessReducer(state).items;
+export const getDevicesReducer = state => state.devices;
+export const getEntities = state => getDevicesReducer(state).entities;
```
mechaffin: getEntities and getItems doesn't have Devices in the name. Should they?

When would we use those instead of getDevices?

src/store/reducers/rulesReducer.js
```diff
@@ -0,0 +1,131 @@
...
+  registerError: { type: 'RULES_REDUCER_ERROR', reducer: errorReducer },
+  isFetching: { multiType: fetchableTypes, reducer: pendingReducer },
+});
+
+export const reducer = { rules: redux.getReducer(initialState) };
+// ========================= Reducers - END
+
+// ========================= Selectors - START
+export const getRulesReducer = state => state.rules;
+export const getEntities = state => getRulesReducer(state).entities;
```
mechaffin: Same comment as I made for getEntities/getItems for devices.

src/components/pages/rules/rules.js
spryor: Not really, at least not here. The propTypes should do that but I haven't done that yet since the props haven't been finalized.

src/components/pages/rules/rules.js
spryor: See comment above.

src/components/shared/refreshBar/refreshBar.js
```diff
@@ -0,0 +1,22 @@
...
+
+import './refreshBar.css';
+
+export const RefreshBar = (props) => (
+  <div className="last-updated-container">
+    {
+      props.isPending || props.time
+        ? <span className="time">
+            <span className="refresh-text">Last refreshed | </span>
+            { !props.isPending ? moment(props.time).format('MM/DD/YYYY, hh:mm:ss') : '--/--/----, --:--:--' }
```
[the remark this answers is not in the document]
spryor: Formatting is a function of the view so I'm leaving it raw in the redux state (easier to compute against). There should be a universal format (and it is coming) but this format matches the current version of RM so I left it.

src/services/httpClient.test.js
```diff
@@ -16,3 +18,23 @@ test('ajax request has timeout parameter', () => {
...
+test("it should support marble tests", marbles((m) => {
+  const { maxRetryAttempts, retryableStatusCodes } = Config;
+
+  const values = {
+    a: 0,
+    '#': { status: 0}
+  };
+
+  const source =  m.cold("-a-|");
+  const asub =     "^--!";
```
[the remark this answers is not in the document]
spryor: Woops, I didn't intend to check this file in.

src/components/pages/rules/rulesGrid/rulesGridConfig.js
```diff
@@ -0,0 +1,98 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+//import SeverityCellRenderer from '../cellRenderers/severityCellRenderer/severityCellRenderer';
```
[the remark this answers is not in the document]
spryor: Forgot about this, I will remove it.

src/store/reducers/appReducer.js
spryor: I put it here because it is a part of the app state. If you look at the current version of RM, the device groups are accessible/changeable from every view. Rules and actions, for example, care about the device groups and currently selected devices groups without references to the devices themselves.

src/store/reducers/devicesReducer.js
spryor: I don't think they should. They are really just there to help inside the devicesReducer (notice how they are being used later in this file).

src/store/reducers/rulesReducer.js
spryor: See above

src/store/reducers/devicesReducer.js
mechaffin: If getEntities and getItems aren't used except internally, we should remove the export from them.

PR #880, opened 2018-03-06
Standardize Date format and Time Zones

molly-moen:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Standardize date format to be locale-specific (ex. for PST this is month/day/year hour:minute:second am/pm), and standardize time zone across dashboard to be local time only.
...

**Checklist:**

- [ ] All tests passed


src/common/utils.js
```diff
@@ -100,10 +103,29 @@ export function copyToClipboard(data) {
   textField.remove();
 };
 
+/*Returns date set to local timezone and in format of locale */
+export function formatLocalDate(date, isUTC)
+{
+  var timezone = moment.tz.guess();
```
spryor: `const` instead of `var`

src/common/utils.js
```diff
@@ -100,10 +103,29 @@ export function copyToClipboard(data) {
   textField.remove();
 };
 
+/*Returns date set to local timezone and in format of locale */
```
spryor: `/*[space]`

src/common/utils.js
```diff
@@ -100,10 +103,29 @@ export function copyToClipboard(data) {
...
+{
+  var timezone = moment.tz.guess();
+  if (isUTC)
+  {
+    date = moment.utc(date);
+  }
+  return moment.tz(date, timezone).toDate().toLocaleString();
+}
+
+/*Returns date set to local timezone and formatted in standard format of YYYY-MM-DDTHH:mm:ss*/
```
spryor: `/*[space]`

src/common/utils.js
```diff
@@ -100,10 +103,29 @@ export function copyToClipboard(data) {
...
+export function convertToLocalTime(date)
```
spryor: `{` on the same line as the `function`

src/common/utils.js
```diff
@@ -100,10 +103,32 @@ export function copyToClipboard(data) {
...
+/* Returns date set to local timezone and in format of locale */
+export function formatLocalDate(date, isUTC, timezone) {
+  if(!timezone) {
+    timezone = moment.tz.guess();
+  }
+  if (isUTC) {
+    date = moment.utc(date);
+  }
+  var localDate = moment.tz(date, timezone);
+  return localDate.format("l") + " " + localDate.format("LTS");
```
molly-moen: "l" = 9/4/1986, "LTS" = 8:30:25 PM, (both are locale-specific). Moment does not have a single locale string for our desired date format.

src/common/utils.js
```diff
@@ -100,10 +103,32 @@ export function copyToClipboard(data) {
   textField.remove();
 };
 
+/* Returns date set to local timezone and in format of locale */
```
avanikp: can you add unit tests for these functions

src/common/utils.js
```diff
@@ -100,10 +103,32 @@ export function copyToClipboard(data) {
...
+export function formatLocalDate(date, isUTC, timezone) {
```
avanikp: perhaps rename to formatToLocalTimeZone ?

avanikp: Can we extract a common method to calculate local time (and handle the case for utc) and have two other functions to get the format - getStandardFormat, getLocaleFormat

src/common/utils.js
```diff
@@ -100,10 +103,32 @@ export function copyToClipboard(data) {
...
+  if(!timezone) {
+    timezone = moment.tz.guess();
+  }
+  if (isUTC) {
```
avanikp: Also add a comment on why we need to handle the case for utc.

package.json
```diff
@@ -12,6 +12,7 @@
     "jquery": "3.2.1",
     "lodash": "4.17.4",
     "moment": "2.18.1",
```
ppathan: Can we look into updating the moment version?

src/common/utils.js
```diff
@@ -100,10 +103,32 @@ export function copyToClipboard(data) {
...
+
 export default {
   isFunction,
   debounce,
   formatDate,
   formatString,
-  getRandomString
+  getRandomString,
+  formatLocalDate,
+  convertToLocalTime
```
ppathan: formatLocalDate and convertToLocalTime show both date and time. In css classname we use ```className="last-refreshed-time"``` It would be great if the method names are consistent either *LocalDate or *LocalTime or *LocalDateTime

src/common/utils.js
molly-moen: added unit tests

ppathan:
@spryor Can you please approve the review if ready.

src/common/utils.test.js
```diff
@@ -0,0 +1,29 @@
...
+
+import { getLocalTimeFormat, getStandardTimeFormat } from './utils'
+
+const utcDateTime = '2018-03-07T12:51:00Z';
+const pacificOffset = '2018-03-07T12:51:00-08:00';
+const timeWithoutZone = '2018-03-07T12:51:00';
+
+describe('Time Zone Conversion', () => {
+    // if time zone is set in string, should work with isUtc set to true and false
+    it('basic convert to -08:00', () => {
```
ppathan: nit: The tests name should read like a sentence.
```Time zone conversion```
```should be able to do basic convesion to -08:00```

File: src/common/utils.js [hunk 12]
```diff
@@ -100,10 +103,41 @@ export function copyToClipboard(data) {
   textField.remove();
 };
 
+/* Returns date set to local timezone and in format of locale */
+export function getLocalTimeFormat(date, isUTC, timezone) {
+  var localDate = getLocalTime(date, isUTC, timezone);
```
spryor: Use `const` instead of `var`

src/common/utils.js
```diff
@@ -100,10 +103,41 @@ export function copyToClipboard(data) {
...
+  return localDate.format("l") + " " + localDate.format("LTS");
+}
+
+/* Returns date set to local timezone and formatted in standard format of YYYY-MM-DDTHH:mm:ss */
+export function getStandardTimeFormat(date, isUTC, timezone) {
+  var localDate = getLocalTime(date, isUTC, timezone);
```
spryor: Use `const` instead of `var`

src/common/utils.js
```diff
@@ -100,10 +103,41 @@ export function copyToClipboard(data) {
...
+  return localDate.format("YYYY-MM-DDTHH:mm:ss");
```
spryor: Two small things 
- Use single quotes instead of double
- Move the actual string to the config file under common/ and import the config here

PR #886, opened 2018-03-10
Add dashboard grid and dashboard alarms panel

spryor:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

- Add the dashboard grid
- Add alarms panel and alarms reducers


src/services/models/configModels.js
```diff
@@ -2,11 +2,10 @@
...
 
-export const toDeviceGroupsModel = (response = {}) => (response.items || []).map(toDeviceGroupModel);
-
-export const toDeviceGroupModel = (device = {}) => reshape(device, {
-  'id': 'id',
-  'displayName': 'displayName',
-  'conditions': 'conditions',
-  'eTag': 'eTag'
-});
+export const toDeviceGroupsModel = (response = {}) => (response.items || [])
```
mechaffin: So by changing this pattern, you're thinking we'll never need one at a time... always a list?

src/components/pages/dashboard/panels/kpis.js
```diff
@@ -0,0 +1,47 @@
...
+import React, { Component } from 'react';
+import { Observable } from 'rxjs';
+
+import { Indicator } from 'components/shared';
+import {
+  Panel,
+  PanelHeader,
+  PanelContent,
+  PanelOverlay
+} from 'components/pages/dashboard/panel';
```
mechaffin: Seems like these will be imported a lot together. Should it have an index.js to simplify?

src/components/pages/dashboard/panels/kpis.js
```diff
@@ -0,0 +1,47 @@
...
+
+  render() {
+    return (
+      <Panel>
+        <PanelHeader>System KPIs</PanelHeader>
+        <PanelContent>
+          Contents
+        </PanelContent>
+        { this.state.isPending && <PanelOverlay><Indicator /></PanelOverlay> }
+      </Panel>
```
mechaffin: These panels with header, overlay and indicator seems like it would be useful outside of the dashboard... since we did something similar when loading grids, should likely use it on details flyouts, etc. Consider moving to components/shared.

src/utilities/svgs.js
```diff
@@ -6,6 +6,7 @@ import ColonIconPath from 'assets/icons/colon.svg';
 import ContosoIconPath from 'assets/icons/contoso.svg';
 import CriticalIconPath from 'assets/icons/critical.svg';
 import DisabledIconPath from 'assets/icons/disabled.svg';
+import ElipsisIconPath from 'assets/icons/elipsis.svg';
```
mechaffin: Check spelling.  It should be with 2 L's... ellipsis.

src/utilities/svgs.js
```diff
@@ -42,6 +43,7 @@ export const svgs = {
   contoso: ContosoIconPath,
   critical: CriticalIconPath,
   disabled: DisabledIconPath,
+  elipsis: ElipsisIconPath,
```
mechaffin: Spelling... ellipsis.

src/components/pages/dashboard/panels/kpis.js
spryor: It is being imported from an index.js file

src/components/pages/dashboard/panels/kpis.js
spryor: These panels are unique to the dashboard

src/services/models/configModels.js
spryor: Just simplifying since we only use it this way right now. If we have a call later on that does shared a model we can break that one out then.

src/utilities/svgs.js
spryor: Woops, good catch

PR #887, opened 2018-03-14
Docker container excessively privileged running root user

zhang-hua:
* Add default non-root user 'pcsuser'
* Install sudo and add 'pcsuser' as sudoer
* Run service as 'pcsuser'

PBI[2211778]

# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

Without explicitly defining a running user in a Dockerfile definition, the root user is utilized by default.
Each respective Dockerfile should explicitly define a user and mandate that user to be the service runner via the USER instruction. Further, the service user should also have a defined GROUP. Without one, the primary group for the user will fall back to the root group.

**Checklist:**

- [x] All tests passed

### DIFF
--- a/scripts/docker/Dockerfile
+++ b/scripts/docker/Dockerfile
@@ -4,6 +4,10 @@ MAINTAINER Devis Lucato (https://github.com/dluc)
 
 LABEL Tags="Azure,IoT,Solutions,React,SPA"
 
+ARG user=pcsuser
+
+RUN adduser -D -h /home/$user -s /bin/bash $user $user
+
 COPY ./ /app
 WORKDIR /app
 
@@ -19,9 +23,14 @@ RUN apk update && apk upgrade \
  && apk add --no-cache nginx \
  && apk del --force --purge alpine-keys apk-tools \
  && rm -rf /var/cache/apk /etc/apk /lib/apk \
- && mkdir /app/logs
+ && mkdir /app/logs \
+ && chown -R $user:$user /app \
+ && mkdir -p /var/lib/nginx /var/cache/nginx /var/tmp/nginx /var/log/nginx \
+ && chown -R $user:$user /var/lib/nginx /var/cache/nginx /var/tmp/nginx /var/log/nginx
 
-EXPOSE 80 443
+EXPOSE 8080 8443
 VOLUME /app/logs
 
 ENTRYPOINT ["/bin/sh", "/app/run.sh"]
+
+USER $user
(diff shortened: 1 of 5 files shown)


scripts/docker/Dockerfile
```diff
@@ -4,6 +4,10 @@ MAINTAINER Devis Lucato (https://github.com/dluc)
 
 LABEL Tags="Azure,IoT,Solutions,React,SPA"
 
+ARG user=pcsuser
```
dluc:
```
ARG user=app
```

scripts/docker/Dockerfile
```diff
@@ -4,6 +4,10 @@ MAINTAINER Devis Lucato (https://github.com/dluc)
...
+
+RUN adduser -D -h /home/$user -s /bin/bash $user $user
```
dluc:
```
RUN groupadd -r $user && useradd --no-log-init -r -g $user $user
```

scripts/docker/content/run.sh
```diff
@@ -3,4 +3,4 @@ set -ex
 
 # serve the app via nginx
 mkdir -p /app/logs
-nginx -g 'daemon off;' -c /app/nginx.conf
```
dluc: what's this change for?

zhang-hua: this is moved into nginx.conf.

scripts/docker/Dockerfile
zhang-hua: 'groupadd' and 'useradd' command are not supported in /bin/sh. I will change them to use 'addgroup' and 'adduser' which works in sh.

ppathan:
@zhang-hua Have you tested this change with hosting your own docker image in vm?

scripts/docker/Dockerfile
```diff
@@ -4,6 +4,10 @@ MAINTAINER Devis Lucato (https://github.com/dluc)
 
 LABEL Tags="Azure,IoT,Solutions,React,SPA"
 
+ARG user=app
+
+RUN addgroup $user && adduser -D -G $user $user
```
dluc: can you sync this with the line of the other PR?

zhang-hua: the command 'groupadd' and 'useradd' are not supported by /bin/sh in Busybox. They are supported by bash shell. I have to change them to 'addgroup' and 'adduser' which are supported by /bin/sh.

zhang-hua:
@ppathan , I did test the PR with private built images before pushing this PR. please also help to validate it if you have time.

scripts/docker/Dockerfile
zhang-hua: the command 'groupadd' and 'useradd' are not supported by /bin/sh in Busybox. They are supported by bash shell. I have to change them to 'addgroup' and 'adduser' which are supported by /bin/sh.

ppathan:
@dluc Any reason we are using bin/sh and not bash for docker image of webui?

PR #888, opened 2018-03-14
Telemetry panel

spryor:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

First pass of the telemetry panel on the dashboard
- add charting library
- get polling working
TODO: Location of calls will be migrated to a single source of truth when all the component data usage is finalized during the dashboard completing process


src/components/pages/dashboard/panels/telemetry/telemetryPanel.js
```diff
@@ -0,0 +1,155 @@
...
+    );
+  }
+
+  setTelemetryKey = telemetryKey => () => this.setState({ telemetryKey });
+
+  render() {
+    const { telemetryKeys, telemetry, telemetryKey } = this.state;
+    return (
+      <Panel>
+        <PanelHeader>Telemetry</PanelHeader>
```
ppathan: Do we need to localize this text? If yes we should move it to a separate file

src/components/pages/dashboard/panels/telemetry/telemetryPanel.js
```diff
@@ -0,0 +1,155 @@
...
+            tooltip: true,
+            yAxisState: 'shared' // Default to all values being on the same axis
+          },
+          chartColors.map(color => ({ color }))
+        );
+      }, 10);
+    }
+  }
+
+  getData() {
```
ppathan: Can you add a comment what this method is doing? Pretty hard to follow

src/components/pages/dashboard/panels/telemetry/telemetryPanel.js
```diff
@@ -0,0 +1,155 @@
...
+} from 'components/pages/dashboard/panel';
+import { joinClasses } from 'utilities';
+
+import './telemetryPanel.css';
+// TODO: find a way to import without the relative path
+import '../../../../../../node_modules/tsiclient/tsiclient.css';
+
+const telemetryChartId = 'device-telemetry-chart';
+
+const chartColors = [
```
ppathan: May be add actual color names in comments for clarity

src/components/pages/dashboard/panels/telemetry/telemetryPanel.js
spryor: Yeah, the calls in this file will be moved to either the redux store or the parent component (`dashboard.js`). I'll document the calls there (and the naming I have in mind will help I think).

src/components/pages/dashboard/panels/telemetry/telemetryPanel.js
spryor: I don't know how helpful that will actually be. We don't have names from design and even then it will be stuff like 'stone' or 'ash'. These colors will be moved to a config later on so the redux store can switch them out when the theme changes.

src/components/pages/dashboard/panels/telemetry/telemetryPanel.js
spryor: Yes it does need to be localized. I haven't done so yet since this is the first pass. I'll do the localization pass when the functionality is working correctly. What do you mean by move to a separate file? You mean move the string? When the string is localized it will be moved to the `public/locales/en/translation.js` file.

ppathan: Yes moving to the locales folder is what i meant. It is ok to do it later

src/components/pages/dashboard/panels/telemetry/telemetryPanel.js
ppathan: If you are already tracking this then it is fine otherwise open a bug.

src/components/pages/dashboard/panels/telemetry/telemetryPanel.js
ppathan: If you are already tracking this then it is fine otherwise open a bug.

ppathan:
@spryor Can you please fix the build failures before check-in?

src/services/telemetryService.js
```diff
@@ -25,12 +31,37 @@ export class TelemetryService {
...
       .map(toAlarmsForRuleModel);
   }
+
+  /** Returns a telemetry events */
+  static getTelemetryByMessages(params = {}) {
+    return HttpClient.get(`${ENDPOINT}messages?${stringify(params)}`)
+      .map(toMessagesModel);
+  }
+
+  static getTelemetryByDeviceIdP1M(deviceIds = '') {
```
avanikp: What is P1m, PT1M, P15M, PT15M ? Can you add comments for these.

src/services/telemetryService.js
```diff
@@ -25,12 +31,37 @@ export class TelemetryService {
...
+    return TelemetryService.getTelemetryByMessages({
+      from: 'NOW-PT1M',
+      to: 'NOW',
+      order: 'desc',
+      devices: encodeURIComponent(deviceIds)
+    });
+  }
+
+  static getTelemetryByDeviceIdP15M(deviceIds = '') {
```
avanikp: Instead of having getTelemetryByDeviceIdP1M and getTelemetryByDeviceIdP15M, can we add a generic method that takes a param for P1M, P15M ?

src/components/pages/dashboard/panels/telemetry/telemetryPanel.js
```diff
@@ -0,0 +1,155 @@
...
+          {
+            grid: false,
+            legend: 'compact',
+            noAnimate, // If the telemetryKey changes, animate
+            tooltip: true,
+            yAxisState: 'shared' // Default to all values being on the same axis
+          },
+          chartColors.map(color => ({ color }))
+        );
+      }, 10);
```
avanikp: suggest making 10 a const

src/components/pages/dashboard/panels/telemetry/telemetryPanel.js
```diff
@@ -0,0 +1,155 @@
...
+    }
+  }
+
+  getData() {
+    this.subscription = TelemetryService.getTelemetryByDeviceIdP15M()
+      .flatMap(response =>
+        Observable.interval(2000)
```
avanikp: nit: declare 2000 as a const

src/services/telemetryService.js
spryor: These are time range constants. These will be documented later. The process I'm currently following is get the basic setup working per dashboard panel then I will be doing a unification pass (making use of duplicated calls etc. by moving the calls into the parent component or the redux store). So this is future work.

src/services/telemetryService.js
spryor: There is a generic method above these `getTelemetryByMessages`. These methods are only temporary though, see comment above.

src/components/pages/dashboard/panels/telemetry/telemetryPanel.js
spryor: Temporary constant so not adding to the config

src/components/pages/dashboard/panels/telemetry/telemetryPanel.js
spryor: I'll take this under consideration but this call will later be wrapped in a component to hide the tsiChart config data. Since this isn't strictly speaking an app constant I would lean away from making an app constant.

PR #892, opened 2018-03-19
Device details - small fixes and styling

mechaffin:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Fixes primarily to styling, but also some in code. Make property and tag headers show up again. Add alarm status column def.

**Checklist:**

- [x] All tests passed


src/components/pages/devices/flyouts/deviceDetails/deviceDetails.js
```diff
@@ -148,7 +150,7 @@ export class DeviceDetails extends Component {
               <Accordion title={t('devices.details.properties.title')} description={t('devices.details.properties.description')}>
                 <Grid>
                   {
-                    (device.properties && device.properties.length) &&
+                    (device.properties && Object.entries(device.properties).length > 0) &&
```
spryor: `Object.keys`

src/components/pages/devices/flyouts/deviceDetails/deviceDetails.js
```diff
@@ -116,7 +117,7 @@ export class DeviceDetails extends Component {
               <Accordion title={t('devices.details.tags.title')} description={t('devices.details.tags.description')}>
                 <Grid>
                   {
-                    (device.tags && device.tags.length) &&
+                    (device.tags && Object.entries(device.tags).length > 0) &&
```
spryor: Use `Object.keys` instead, that is the convention for object size and you don't need the values here anyway.

src/components/pages/devices/flyouts/deviceDetails/accordion.scss
```diff
@@ -21,15 +21,16 @@
     }
 
     .accordion-toggle-btn {
+      cursor: pointer;
       background-color: transparent;
+      border: 0;
       @include rem-fallback(padding, 0px);
```
spryor: If you are setting the value to 0 you don't need to use REMs or have a px fallback. Just use `padding: 0; margin: 0;`

src/components/pages/devices/flyouts/deviceDetails/accordion.js
```diff
@@ -33,15 +33,19 @@ export class Accordion extends Component {
...
+          </button>
         </SectionHeader>
 
-        <div className="section-content" style={displayContent}>
-          {
-            description &&
-            <SectionDesc>{description}</SectionDesc>
-          }
-          {children}
+        <div className="section-content-container">
```
spryor: I don't see why there are two divs  here. Seems like there should just be one.

src/components/pages/devices/flyouts/deviceDetails/deviceDetails.js
```diff
@@ -116,7 +117,7 @@ export class DeviceDetails extends Component {
               <Accordion title={t('devices.details.tags.title')} description={t('devices.details.tags.description')}>
                 <Grid>
                   {
-                    (device.tags && device.tags.length) &&
+                    (device.tags && Object.keys(device.tags).length > 0) &&
```
spryor: I would simplify this and the below code a bit. Before the return, call `const tags = Object.keys(device.tags || {})` and save the result. Then you can use the result here and to loop through the results later on.

src/components/pages/devices/flyouts/deviceDetails/accordion.js
```diff
@@ -33,15 +33,19 @@ export class Accordion extends Component {
...
+          <div className="section-content" style={displayContent}>
+            {
+              description &&
```
spryor: make 44 and 45 one line

src/components/pages/devices/flyouts/deviceDetails/accordion.js
mechaffin: Because I needed to be able to hide the actual contents (section-content), but also maintain the spacing (using section-content-container).

src/components/pages/devices/flyouts/deviceDetails/deviceDetails.js
```diff
@@ -116,14 +119,14 @@ export class DeviceDetails extends Component {
               <Accordion title={t('devices.details.tags.title')} description={t('devices.details.tags.description')}>
                 <Grid>
                   {
-                    (device.tags && device.tags.length) &&
+                    (tags.length > 0) &&
```
spryor: Didn't notice this a second ago but wrap the grid itself with the `(tags.length > 0)` check and if nothing is there display some text saying there are no tags. Same the for content below here.

src/components/pages/devices/flyouts/deviceDetails/deviceDetails.js
```diff
@@ -114,55 +118,71 @@ export class DeviceDetails extends Component {
...
-                    ((device.methods || '').split(',') || []).map((methodName, idx) =>
-                      <Row key={idx}>
-                        <Cell>{methodName}</Cell>
-                      </Row>
-                    )
-                  }
-                </Grid>
+                {
+                  (methods.length === 0) &&
+                  <div>{t('devices.details.methods.noneExist')}</div>
```
spryor: I don't think the div is needed. Did it cause an error without it?

src/components/pages/devices/flyouts/deviceDetails/deviceDetails.js
```diff
@@ -114,55 +118,71 @@ export class DeviceDetails extends Component {
...
+                }
               </Accordion>
 
               <Accordion title={t('devices.details.properties.title')} description={t('devices.details.properties.description')}>
-                <Grid>
-                  {
-                    (device.properties && device.properties.length) &&
+                {
+                  (properties.length === 0) &&
+                  <div>{t('devices.details.properties.noneExist')}</div>
```
spryor: Same question about the div

PR #897, opened 2018-03-23
Device Delete Flyout, Btn styling.

mechaffin:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

Device Details Flyout implemented.
Also, added styling for shared Btn controls (mostly) per the redlines.

**Checklist:**

- [x] All tests passed


src/components/pages/devices/flyouts/deviceDelete/deviceDelete.js
```diff
@@ -0,0 +1,135 @@
...
+
+    this.subscription = Observable.from(this.state.physicalDevices)
+      .flatMap(({ id }) =>
+        IoTHubManagerService.deleteDevice(id)
+          .map(() => id)
+      )
+      .subscribe(
+        next => {
+          this.setState({ successCount: this.state.successCount + 1 });
+          return this.props.deleteDevice(next);
```
spryor: Not need to return

src/components/pages/devices/flyouts/deviceDelete/deviceDelete.js
```diff
@@ -0,0 +1,135 @@
...
+        },
+        error => this.setState({ error, isPending: false, changesApplied: true }),
+        () => this.setState({ isPending: false, changesApplied: true })
+      );
+
+  }
+
+  render() {
+    const { t, onClose } = this.props;
+    const { physicalDevices, containsSimulatedDevices, isPending, error, successCount, changesApplied } = this.state;
```
spryor: Break this into multiple lines
```
const {
    physicalDevices,
    containsSimulatedDevices,
   ...
```

src/components/pages/devices/flyouts/deviceDelete/deviceDelete.js
```diff
@@ -0,0 +1,135 @@
...
+        </FlyoutHeader>
+        <FlyoutContent>
+          <div className="device-delete-container">
+            <div className="device-delete-header">{t('devices.delete.header')}</div>
+            <div className="device-delete-descr">{t('devices.delete.description')}</div>
+            {
+              containsSimulatedDevices &&
+              <div className="simulated-device-selected">
+                <Svg path={svgs.infoBubble} className="info-icon" />
+                <div>{t('devices.delete.simulatedNotSupported')}</div>
```
spryor: I don't think you need the div

src/components/pages/devices/flyouts/deviceDelete/deviceDelete.js
```diff
@@ -0,0 +1,135 @@
...
+              </div>
+            }
+            <div className="device-delete-summary-container">
+              <div>
```
spryor: empty div

src/components/pages/devices/flyouts/deviceDelete/deviceDelete.js
```diff
@@ -0,0 +1,135 @@
...
+                Summary
+              </div>
+              <div className="device-delete-affected">
+                <div className="device-delete-count">{summaryCount}</div>
+                <div>{summaryMessage}</div>
```
spryor: empty div

src/components/shared/forms/styles/btn.scss
```diff
@@ -34,24 +36,76 @@ $btnIconSize: 14px;
...
+        svg { fill: themed('colorBtnSvgFillDisabled'); }
+      }
     }
 
+    &.btn-primary {
+      color: themed('colorBtnPrimaryText');
+      background-color: themed('colorBtnPrimaryBackground');
+      .btn-icon svg { fill: themed('colorBtnPrimarySvgFill'); }
+
+      &:focus {
```
spryor: online

src/components/shared/forms/styles/btn.scss
```diff
@@ -34,24 +36,76 @@ $btnIconSize: 14px;
...
+      .btn-icon {
+        display: none;
+
+        svg { fill: themed('colorBtnSvgFillDisabled'); }
+      }
     }
 
+    &.btn-primary {
+      color: themed('colorBtnPrimaryText');
+      background-color: themed('colorBtnPrimaryBackground');
```
spryor: add newline

src/components/shared/forms/styles/btn.scss
```diff
@@ -34,24 +36,76 @@ $btnIconSize: 14px;
...
+      color: themed('colorBtnTextHover');
+      background-color: themed('colorBtnBackgroundHover');
+
+      .btn-icon svg { fill: themed('colorBtnSvgFillHover'); }
+    }
+
+    &:active {
+      color: themed('colorBtnTextActive');
+      background-color: themed('colorBtnBackgroundActive');
+      border: 1px transparent dashed;
```
spryor: If the property isn't changing with the theme, it shouldn't be inside the themify.

src/components/shared/forms/styles/btn.scss
```diff
@@ -34,24 +36,76 @@ $btnIconSize: 14px;
...
+      }
     }
 
+    &.btn-primary {
+      color: themed('colorBtnPrimaryText');
+      background-color: themed('colorBtnPrimaryBackground');
+      .btn-icon svg { fill: themed('colorBtnPrimarySvgFill'); }
+
+      &:focus {
+        border: 1px themed('colorBtnPrimaryBorderFocus') dashed;
```
spryor: The focus should probably be an `outline`

src/services/httpClient.js
```diff
@@ -69,7 +69,7 @@ export class HttpClient {
     const request = HttpClient.createAjaxRequest({ ...options, url }, withAuth);
     return Observable.ajax(request)
       // If success, extract the response object and enforce camelCase keys
-      .map(({ response }) => camelcaseKeys(response, { deep: true }))
+      .map(({ response }) => response && camelcaseKeys(response, { deep: true }))
```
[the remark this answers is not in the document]
mechaffin: Correct. Delete didn't return a response body at all.

src/store/reducers/devicesReducer.js
```diff
@@ -66,6 +66,15 @@ const updateDevicesReducer = (state, { payload, fromAction }) => {
   });
 };
 
+const deleteDeviceReducer = (state, { payload }) => {
+  const itemIdx = state.items.findIndex(element => element === payload);
+  return update(state, {
+    entities: { $unset: [payload] },
+    items: { $splice: [[itemIdx, 1]] },
+    lastUpdated: { $set: moment() }
```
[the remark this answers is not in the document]
mechaffin: OK, I was unsure of this as well.  I'm totally OK with it being the server sync time.

src/styles/_themes.scss
```diff
@@ -38,6 +38,44 @@ $themes: (
     colorCellRendererTextHighlight: $colorWhite,
     // Grid - END
 
+    //Btn - START
+    colorBtnBorderFocus: $colorSmoke,
+
+    colorBtnText: $colorWhite,
+    colorBtnBackground: transparent,
+    colorBtnSvgFill: $colorSmoke,
```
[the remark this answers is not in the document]
mechaffin: Unfortunately, I cannot just use the same color. In some cases, they are different.

src/components/pages/devices/flyouts/deviceDelete/deviceDelete.js
mechaffin: This changes to a "summarySection" in my next PR... so will leave as is for now.

src/components/pages/devices/devicesGrid/devicesGrid.js
```diff
@@ -86,11 +115,16 @@ export class DevicesGrid extends Component {
...
+      onSoftSelectChange: this.onSoftSelectChange,
       onHardSelectChange: this.onHardSelectChange,
       onGridReady: this.onGridReady
     };
-    return (
-      <PcsGrid {...gridProps} />
-    );
+    const openFlyout = (this.state.openFlyoutName === 'delete')
+      ? <DeviceDeleteContainer key="device-flyout-key" onClose={this.closeFlyout} devices={this.deviceGridApi.getSelectedRows()} />
+      : undefined;
```
spryor: I typically return `null` instead of `undefined`. `undefined` will still work off course but I prefer the semantics of  `null`. The reason is that `null` is a value that is a "non-value" where as `undefined` says there is not value at all. It seems more reasonable to have an array where all values are `defined` but some are `null` rather than some values are `undefined`.

src/components/pages/devices/flyouts/deviceDelete/deviceDelete.js
```diff
@@ -0,0 +1,142 @@
...
+      this.populateDevicesState(nextProps.devices);
+    }
+  }
+
+  componentWillUnmount() {
+    if (this.subscription) this.subscription.unsubscribe();
+  }
+
+  populateDevicesState = (devices) => {
+    const deviceArr = devices || [];
```
spryor: You don't need this since you can set the default value in the method signature.
`populateDevicesState = (devices = []) => {`

src/components/pages/devices/flyouts/deviceDelete/deviceDelete.js
```diff
@@ -0,0 +1,142 @@
...
+            <div className="device-delete-descr">{t('devices.delete.description')}</div>
+            {
+              containsSimulatedDevices &&
+              <div className="simulated-device-selected">
+                <Svg path={svgs.infoBubble} className="info-icon" />
+                {t('devices.delete.simulatedNotSupported')}
+              </div>
+            }
+            <div className="device-delete-summary-container">
+              <div>
```
spryor: Why is this vanilla div here?

src/components/shared/forms/btn.js
```diff
@@ -9,9 +9,9 @@ import { joinClasses } from 'utilities';
 import './styles/btn.css';
 
 export const Btn = (props) => {
-  const {svg, children, className, ...btnProps } = props;
+  const { svg, children, className, primary, ...btnProps } = props;
   return (
-    <button type="button" {...btnProps} className={joinClasses('btn', className)}>
+    <button type="button" {...btnProps} className={joinClasses('btn', className, primary ? 'btn-primary' : 'btn-secondary')}>
```
spryor: I thought there were three versions of buttons?

src/services/httpClient.js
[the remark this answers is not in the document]
spryor: I see. Instead of having the stream return an undefined value let's do this instead.

`camelcaseKeys((response || {}), { deep: true })`

That way we will just get an empty object instead of nothing.

src/services/iotHubManagerService.js
```diff
@@ -15,4 +15,9 @@ export class IoTHubManagerService {
     return HttpClient.get(`${ENDPOINT}devices?query=${query}`)
       .map(toDevicesModel);
   }
+
+  /** Deletes a device */
+  static deleteDevice(id) {
+    return HttpClient.delete(`${ENDPOINT}devices/${id}`);
```
spryor: Let's add a map here to create a synthetic response object.

```
return HttpClient.delete(`${ENDPOINT}devices/${id}`)
    .map(() => ({ deletedDeviceId: id }))
```

This will simplify the code in the flyout an other locations where we may need to delete a device.

src/store/reducers/devicesReducer.js
```diff
@@ -66,6 +66,14 @@ const updateDevicesReducer = (state, { payload, fromAction }) => {
   });
 };
 
+const deleteDeviceReducer = (state, { payload }) => {
+  const itemIdx = state.items.findIndex(deviceId => deviceId === payload);
```
spryor: Since items is an array of deviceId's you can use `indexOf` instead.

```
const itemIdx = state.items.indexOf(payload);
```

src/components/pages/devices/devicesGrid/devicesGrid.js
```diff
@@ -60,6 +76,19 @@ export class DevicesGrid extends Component {
     }
   };
 
+  /**
+   * Handles soft select props method
+   *
+   * @param {Array} selectedDevices A list of currently selected devices
+   */
+  onSoftSelectChange = (rowEventData, rowEvent) => {
```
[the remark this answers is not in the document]
mechaffin: Yes, those are the names based on the underlying component... PcsGrid. See line 99-->           onSoftSelectChange(rowEvent.data, rowEvent);

src/components/pages/devices/devicesGrid/devicesGrid.js
```diff
@@ -60,6 +76,19 @@ export class DevicesGrid extends Component {
...
+    const { onSoftSelectChange } = this.props;
+    this.setState(closedFlyoutState);
```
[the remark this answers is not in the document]
mechaffin: The current UI only has 1 open at a time... so I was following that.  There are likely odd cases where it doesn't make sense which flyout might be underneath.

src/components/shared/forms/btn.js
mechaffin: Nope.  Just two... but with lots of states for each... hover, active, focus, disabled.

src/components/pages/devices/flyouts/deviceDelete/deviceDelete.js
```diff
@@ -0,0 +1,141 @@
...
+              error &&
+              <div className="device-delete-error">
+                <ErrorMsg>{error}</ErrorMsg>
+              </div>
+            }
+            {
+              !changesApplied &&
+              <BtnToolbar className="tools-preApply">
+                {this.state.isPending && <Indicator />}
+                <Btn svg={svgs.trash} primary={true} disabled={isPending || physicalDevices.length === 0} onClick={this.deleteDevices}>{t('devices.delete.apply')}</Btn>
```
spryor: I would do `primary={true}` for the buttons. It should either be a single attribute with an implicit true value or it should be something like `btnType="primary"`

src/components/shared/forms/styles/btn.scss
```diff
@@ -33,25 +35,69 @@ $btnIconSize: 14px;
...
+
   @include themify($themes) {
-    color: themed('colorHighest');
+    color: themed('colorBtnText');
+    background-color: themed('colorBtnBackground');
+
+    .btn-icon svg { fill: themed('colorBtnSvgFill'); }
 
-    .btn-icon svg { fill: themed('colorHighest'); }
+    &:focus {
```
spryor: single line

src/components/pages/devices/flyouts/deviceDelete/deviceDelete.js
mechaffin: I thought this was how you told me to do it to begin with.  Otherwise, tell me this 4 days ago.

mechaffin: I'm not sure I agree with an essentially free form string value like btnType would be.  Its really no better than using className and joining them under the covers.

src/components/shared/forms/styles/btn.scss
```diff
@@ -3,28 +3,28 @@
...
+$btnMinWidth: 120px;
 $btnIconSize: 14px;
 .btn {
   display: flex;
   flex-flow: row nowrap;
   flex-shrink: 0;
   align-items: center;
   background-color: transparent;
-  border: none;
+  border: 0px;
```
spryor: no need for px, just use 0

PR #902, opened 2018-03-24
Add context menu container

spryor:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

Add first cut of the context menu. Import the page content container into the page components and return an array when the context menu is needed.

TODO: Update the exports in the components/app folder to be easier to import (remove defaults)
TODO: Add custom styling to Btns when used inside the context menu


src/components/pages/devices/devices.js
```diff
@@ -38,19 +42,26 @@ export class Devices extends Component {
...
       softSelectId: this.state.selectedDeviceId,
       getSoftSelectId: this.getSoftSelectId,
       t: this.props.t
     };
-    return (
-      <div className="devices-container">
+    return [
+      <ContextMenu>
+        { this.state.contextBtns }
+        <Btn svg={svgs.plus}>New device</Btn>
```
spryor: Needs to be translated, this is only here as an example

ppathan: Add a TODO

PR #906, opened 2018-04-03
Translate maintenance page

spryor:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

Translate the strings in the maintenance page
Add a time renderer to avoid naming confusion

### DIFF
--- a/src/app.config.js
+++ b/src/app.config.js
@@ -14,9 +14,12 @@ const Config = {
   retryWaitTime: 2000, // On retryable error, retry after 2s
   retryableStatusCodes: new Set([ 0, 502, 503 ]),
   paginationPageSize: 50,
+  smallGridPageSize: 8,
   clickDebounceTime: 180, // ms
   dashboardRefreshInterval: 15000, // 15 seconds
-  telemetryRefreshInterval: 1000 // 1 seconds
+  telemetryRefreshInterval: 1000, // 1 seconds
+  emptyValue: '--',
+  maxTopAlarms: 5
 };
 
 export default Config;
--- a/src/app.config.js
+++ b/src/app.config.js
@@ -1,5 +1,5 @@
 // Copyright (c) Microsoft. All rights reserved.
-const baseUrl = process.env.REACT_APP_BASE_SERVICE_URL || 'https://maryellen-0226.azurewebsites.net';
+const baseUrl = process.env.REACT_APP_BASE_SERVICE_URL || '';
 
 const Config = {
   // TEMP: Base service urls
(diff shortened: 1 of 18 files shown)


src/app.config.js
```diff
@@ -1,5 +1,5 @@
 // Copyright (c) Microsoft. All rights reserved.
-const baseUrl = process.env.REACT_APP_BASE_SERVICE_URL || '';
+const baseUrl = process.env.REACT_APP_BASE_SERVICE_URL || 'https://maryellen-0226.azurewebsites.net';
```
ppathan: Remove: 'https://maryellen-0226.azurewebsites.net' Does the .env value doesn't work?

src/utilities/methods.js
```diff
@@ -49,3 +49,16 @@ export const isDef = (val) => typeof val !== 'undefined';
 
 /** Return a generic config value if the value is undefined */
 export const renderUndefined = (value) => !isDef(value) ? Config.emptyValue : value;
+
+/** Converts a job status code to a translated string equivalent */
+export const getStatusCode = (code, t) => {
+  switch (code) {
+    case 1: return t('maintenance.jobStatus.enqueued');
```
ppathan: In some place you use t('some string variable in translate file') in some case just [this](https://github.com/Azure/pcs-remote-monitoring-webui/pull/906/files#diff-ffdbd293f00d3299797b078c0f67cf56R12) what's the difference

src/app.config.js
spryor: woops

src/utilities/methods.js
spryor: In the column definitions we don't have access to the `t` method yet (since it comes through props). So what we do is define the columns with the string path in the translation file and then translate it when we need to using the `translateColumnDefs` helper method in the utilities.

src/components/pages/maintenance/grids/jobStatusGrid.js
```diff
@@ -2,51 +2,38 @@
...
+    headerName: 'maintenance.jobStatusGrid.deviceIdAffected',
     field: 'devices',
     valueFormatter: ({ value = [] }) => value.map(({ deviceId }) => deviceId).join(', ')
   },
   {
-    headerName: 'Last Return Message', // TODO: Translate
+    headerName: 'maintenance.jobStatusGrid.lastReturnMsg',
     field: 'status',
-    valueFormatter: ({ value, data = {} }) => `${data.methodName} ${getStatusCode(value)}` // getStatusCode(value)
+    valueFormatter: ({ value, data = {}, context: { t } }) => `${data.methodName} ${getStatusCode(value, t)}` // getStatusCode(value)
```
mechaffin: Remove comment at end of line.

PR #913, opened 2018-04-06
Add map alerts

spryor:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

Label devices on the map with normal, warning, or critical status
Setup a fake service call to get the Azure map key


src/components/pages/dashboard/panels/map/mapPanel.js
```diff
@@ -14,73 +15,133 @@ import {
 
 import './mapPanel.css';
 
-const atlas = window.atlas;
-const deviceLayer = 'devices-layer';
+const AzureMaps = window.atlas;
```
ppathan: azureMaps?

src/components/pages/dashboard/panels/map/mapPanel.js
```diff
@@ -14,73 +15,133 @@ import {
...
-      });
-
-      this.calculatePins(this.props, true);
+    if (!this.map && this.props.azureMapsKey) {
+      this.initializeMap(this.props.azureMapsKey);
     }
+    this.calculatePins(this.props, true);
+    // this.map.addEventListener("click", deviceLayer, ({ features: [ pin ] }) => {
+    //   console.log("Clicked on ", pin);
+    // });
```
ppathan: remove if not needed.

src/components/pages/dashboard/panels/map/mapPanel.js
```diff
@@ -1,6 +1,7 @@
 // Copyright (c) Microsoft. All rights reserved.
 
 import React, { Component } from 'react';
+import update from 'immutability-helper';
```
ppathan: Other imports start with upper case letter

src/components/pages/dashboard/panels/map/mapPanel.js
```diff
@@ -14,73 +15,133 @@ import {
...
-      const neLon = Math.max.apply(null, lons);
-      const neLat = Math.max.apply(null, lats);
-
-      // Zoom the map to the bounding box of the devices
-      this.map.setCameraBounds({
-        bounds: [swLon, swLat, neLon, neLat],
-        padding: 50
-      });
+    const prevDeviceIds = Object.keys(this.props.devices);
+    const boundZoomToDevices =
```
ppathan: Can you add a comment about this condition check? It would be helpful for future ref.

src/components/pages/dashboard/panels/map/mapPanel.js
```diff
@@ -14,73 +15,133 @@ import {
...
+
+    if (this.map && deviceIds.length > 0) {
+      const geoLocatedDevices = this.extractGeoLocatedDevices(props.devices);
+
+      const { normal, warning, critical }  = this.devicesToPins(geoLocatedDevices, props.devicesInAlarm)
+
+      if (this.map && (normal.length || warning.length || critical.length)) {
+        this.map.addPins(normal, { name: nominalDeviceLayer, overwrite: true });
+        this.map.addPins(warning, { name: warningDevicesLayer, overwrite: true });
+        this.map.addPins(critical, { name: criticalDevicesLayer, overwrite: true });
```
ppathan: The || condition will execute even when only one variable has some value. Do you want to add pins if lets say `warning.length == 0` and `normal.length != 0`

src/services/configService.js
```diff
@@ -16,6 +18,12 @@ export class ConfigService {
       .map(toDeviceGroupsModel);
   }
 
+  /** Returns the azure map key for the account */
+  static getAzureMapKey() {
+    return Observable.of('') // TODO: Implement the real service call
+      .delay(2000);
```
ppathan: You can implement the call. The service call is not going to change from before.

src/components/pages/dashboard/panels/map/mapPanel.js
spryor: I'm using this as an alias instead of the more generic "atlas"

src/components/pages/dashboard/panels/map/mapPanel.js
spryor: This is importing a function. Other imports are importing classes and react components.

src/services/configService.js
spryor: I can add it in it's own PR.

src/components/pages/dashboard/panels/map/mapPanel.js
spryor: Actually, we don't need this check at all. I removed it.

src/components/pages/dashboard/panels/map/mapPanel.js
spryor: Done

src/components/pages/dashboard/panels/map/mapPanel.js
spryor: Woops, removed.

src/components/pages/dashboard/panels/map/mapPanel.js
ppathan: That will be confusing for the new developers to follow. Do we have linting that can validate against these kind of requirements?

spryor: The idea is that it should be less confusing. Atlas doesn't really tell you what the object is. AzureMaps does. How would linting help here?

PR #919, opened 2018-04-11
New Device Flyout - Simulated Devices

mechaffin:
# Type of change? 

- [ ] Bug fix
- [X] New feature
- [ ] Enhancement

New Device flyout supports simulated devices.
Also, added validation.

**Checklist:**

- [x] All tests passed


src/components/pages/devices/flyouts/deviceNew/deviceNew.container.js
```diff
@@ -3,11 +3,16 @@
...
+import {
+  epics as devicesEpics,
+  redux as devicesRedux
+} from 'store/reducers/devicesReducer';
 
 // Wrap the dispatch method
 const mapDispatchToProps = dispatch => ({
-  insertDevice: device => dispatch(devicesRedux.actions.insertDevice(device))
+  insertDevice: device => dispatch(devicesRedux.actions.insertDevice(device)),
+  //fetchDevices: () => dispatch(devicesEpics.actions.refreshDevices())
```
maple10001: Please remove commented code.

src/components/pages/devices/flyouts/deviceNew/deviceNew.js
```diff
@@ -129,16 +129,24 @@ const ProvisionedDevice = ({ device, t }) => {
 export class DeviceNew extends LinkedComponent {
   constructor(props) {
     super(props);
+
+    const { t } = this.props;
```
maple10001: Please remove this line, if `t` is declared on line 297.

mechaffin: That's a completely different function. So its needed in both places unless I fully qualify it everywhere.

spryor: `t` doesn't need to be in the constructor. The constructor is only fired once but the `t` function needs to fire if the props change.

src/services/deviceSimulationService.js
```diff
@@ -13,18 +15,56 @@ const SIMULATION_ID = Config.simulationId;
...
+  static toggleSimulation(Etag, Enabled) {
+    return HttpClient.patch(`${ENDPOINT}simulations/${SIMULATION_ID}`, { Etag, Enabled })
+      .map(toDeviceSimulationModel);
+  }
+
+  /**
+   * Gets the simulated device models, increments the given one, then updates on the server
+   */
+  static incrementSimulatedDeviceModel(deviceModelId, increment) {
+    return this.getSimulatedDevices()
```
spryor: Is using `this` working? I don't think it should be. I thought you have to refer to the class directly to access the static methods. E.g. `DeviceSimulationService.updateSimulatedDevices`

mechaffin: Yes, it is working.  Because, um, I usually make it a practice to ensure the code is functioning prior to checking it in.  Hahaha.
But yes, I'll change it.

src/components/pages/devices/flyouts/deviceNew/deviceNew.js
mechaffin: Huh? t is used in all the validation messages... that are set up within the constructor. See lines 162-193.

src/services/deviceSimulationService.js
timlaverty: Definitely not as readable even if it's working - 'this' implies an instance.

src/components/pages/devices/flyouts/deviceNew/deviceNew.js
```diff
@@ -159,15 +163,17 @@ export class DeviceNew extends LinkedComponent {
...
-    this.deviceIdLink = this.formDataLink.forkTo('deviceId');
+      .check(Validator.notEmpty, () => this.props.t('devices.flyouts.new.validation.required'))
+      .check(num => num > 0, () => this.props.t('devices.flyouts.new.validation.greaterThanZero'));
 
     this.isGenerateIdLink = this.formDataLink.forkTo('isGenerateId')
       .map(stringToBoolean);
 
-    this.deviceModelLink = this.formDataLink.forkTo('deviceModel');
+    this.deviceIdLink = this.formDataLink.forkTo('deviceId')
+      .check(devId => (!this.deviceTypeLink.value && !this.isGenerateIdLink.value ? Validator.notEmpty(devId) : true), () => this.props.t('devices.flyouts.new.validation.required'));
```
spryor: For the really long ones maybe break it to one parameter per line.

```
.check(
  devId => (!this.deviceTypeLink.value && !this.isGenerateIdLink.value ? Validator.notEmpty(devId) : true),
  () => this.props.t('devices.flyouts.new.validation.required')
)
```

src/components/pages/devices/flyouts/deviceNew/deviceNew.js
```diff
@@ -206,6 +219,16 @@ export class DeviceNew extends LinkedComponent {
     return true;
   }
 
+  fetchDeviceModelOptions = () => {
+    this.setState({ deviceModels: { isPending: true } });
+
+    this.deviceModelSubscription = DeviceSimulationService.getDeviceModelSelectOptions()
```
spryor: I believe device models are an app level setting. They don't change much, right? They are more like config settings from the server. Loading the device models should probably be placed in the app reducers/epics.

src/store/reducers/deviceSimulationReducer.js
```diff
@@ -32,6 +32,14 @@ export const epics = createEpicScenario({
       DeviceSimulationService.toggleSimulation(fromAction.payload.etag, fromAction.payload.enabled)
         .map(toActionCreator(redux.actions.getSimulationStatus, fromAction))
         .catch(handleError(fromAction))
+  },
+
+  /** Loads the options for Device Models */
+  fetchSimulationDeviceModelOptions: {
+    type: 'SIMULATION_DEVICE_MODEL_OPTIONS_FETCH',
+    epic: (fromAction) => DeviceSimulationService.getDeviceModelSelectOptions()
+      .map(toActionCreator(redux.actions.getDeviceModelOptions, fromAction))
```
spryor: Should probably follow the indentation and new line pattern from above.
```
epic: (fromAction) => 
  DeviceSimulationService.getDeviceModelSelectOptions()
    .map(toActionCreator(redux.actions.getDeviceModelOptions, fromAction))
```

src/components/pages/devices/flyouts/deviceNew/deviceNew.js
```diff
@@ -159,15 +162,23 @@ export class DeviceNew extends LinkedComponent {
...
-    this.deviceModelLink = this.formDataLink.forkTo('deviceModel');
+    this.deviceIdLink = this.formDataLink.forkTo('deviceId')
+      .check(
+        devId => (!this.deviceTypeLink.value && !this.isGenerateIdLink.value ? Validator.notEmpty(devId) : true),
+        () => this.props.t('devices.flyouts.new.validation.required')
+      );
+
+    this.deviceModelLink = this.formDataLink.forkTo('deviceModel')
+      .check(
+        devModel => (this.deviceTypeLink.value ? Validator.notEmpty(devModel) : true),
```
spryor: No real need to wrap these ternary operator with `()` unless there are several conditions that need to be clarified.

PR #920, opened 2018-04-12
Fix failing build

spryor:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Fix issues in the tests and the camelcaseKeys library causing the build to fail


src/components/pages/dashboard/dashboard.test.js
```diff
@@ -1,17 +1,29 @@
 // Copyright (c) Microsoft. All rights reserved.
 
 import React from 'react';
-import { mount } from 'enzyme';
+import { shallow } from 'enzyme';
```
ppathan: With `shallow` we are not loading the whole component. If we want to write integration test, will we be able to use `mount` and mock out the service layer?

package.json
```diff
@@ -6,7 +6,7 @@
     "adal-angular": "1.0.15",
     "ag-grid": "16.0.1",
     "ag-grid-react": "16.0.0",
-    "camelcase-keys": "4.2.0",
+    "camelcase-object": "1.0.0",
```
ppathan: So was build failure fixed just by moving to `camelcase-object`?

src/components/pages/dashboard/dashboard.test.js
spryor: With shallow we load the entire component just not it's children. For integration we can use mount. If you look at the [main app test](https://github.com/Azure/pcs-remote-monitoring-webui/blob/buildFixes/src/components/app/app.test.js) it is still using `mount` to test the deep component integration and routing.

package.json
spryor: Yep. `camelcase-keys` was written to be used with node, not in the browser. So babel was freaking out because of es5 issues. `camelcase-object` works correctly in the browser.

PR #938, opened 2018-04-20
Update casing for the lat and long properties

spryor:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

The casing for the properties response object was changed in a recent PR and the map panel wasn't updated.


src/components/pages/dashboard/panels/map/mapPanel.js
```diff
@@ -115,9 +115,12 @@ export class MapPanel extends Component {
         (acc, device) => {
           const devicePin = deviceToMapPin(device);
           const category = device.id in devicesInAlarm ? devicesInAlarm[device.id].severity : 'normal';
-          return update(acc, {
-            [category]: { $push: [devicePin] }
-          });
+          if (category === 'normal' || category === 'warning' || category === 'critical') {
```
ppathan: We should have normal, warning and critical as enums or const

PR #939, opened 2018-04-20
Rule model changes

isaac73:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

Update in Rule Model
...

**Checklist:**

- [x] All tests passed


src/components/pages/rules/flyouts/ruleEditor/ruleEditor.js
```diff
@@ -95,36 +99,35 @@ export class RuleEditor extends LinkedComponent {
...
+      this.calculationLink
     ].every(link => !link.error);
   }
 
   apply = (event) => {
     event.preventDefault();
     const { formData } = this.state;
     const { onClose, insertRule, updateRule } = this.props;
-    console.log('TODO: Handle the form submission');
+    if (formData.calculation === calculations[1]) this.setState({ formData: { ...this.state.formData, timePeriod: '' } })
```
spryor: No need to update the state here. This could cause some async issues. Just update a local version of the state formData.

src/components/pages/rules/flyouts/ruleEditor/ruleEditor.js
```diff
@@ -95,36 +99,35 @@ export class RuleEditor extends LinkedComponent {
...
       } else { // If rule object doesn't exist then create a new rule
         this.subscription = TelemetryService.createRule(toNewRuleRequestModel(formData))
           .subscribe(
             (createdRule) => {
               insertRule(createdRule);
-              this.setState({ isPending: false });
               onClose();
             },
-            error => this.setState({ error, isPending: false })
+            error => this.setState({ error })
```
spryor: You'll still need to know the pending status to alert the user that their is an action happening. Why are you removing it?

isaac73: That was not in the redline or in previous version

spryor: The previous version is wrong. If it isn't in the redline it probably should be. You should ask about it.

src/components/pages/rules/flyouts/ruleEditor/ruleEditor.js
```diff
@@ -188,20 +191,28 @@ export class RuleEditor extends LinkedComponent {
     this.deviceGroupLink = this.formDataLink.forkTo('groupId')
       .map(({ value }) => value)
       .withValidator(requiredValidator);
+    this.calculationLink = this.formDataLink.forkTo('calculation').map(({ value }) => value).withValidator(requiredValidator);
+    this.timePeriodLink = this.formDataLink.forkTo('timePeriod')
+      .map(({ value }) => value)
+      .check(
+        timePeriod => this.calculationLink.value === calculations[0] ? Validator.notEmpty(timePeriod) : true,
+        () => this.props.t('rules.flyouts.ruleEditor.validation.required')
```
spryor: Function isn't needed inside the render function.

PR #940, opened 2018-04-23
Device Tags Flyout

mechaffin:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

Device Tags Flyout - mostly working

**Checklist:**

- [ ] All tests passed


src/components/pages/devices/flyouts/deviceJobs/deviceJobTags.js
```diff
@@ -1,20 +1,311 @@
...
+      {
+        value: TagJobConstants.stringType,
+        label: t('devices.flyouts.jobs.tags.typeString')
+      }
+    ];
+
+    // Link these values in render because they need to update based on component state
+    const tagLinks = this.tagsLink.getLinkedChildren(tagLink => {
+      const name = tagLink.forkTo('name')
+        .check(Validator.notEmpty, () => this.props.t('devices.flyouts.jobs.validation.required'));
```
maple10001: May use `.check(Validator.notEmpty, this.props.t('devices.flyouts.jobs.validation.required'));` instead of `.check(Validator.notEmpty, () => this.props.t('devices.flyouts.jobs.validation.required'));`

src/components/pages/devices/flyouts/deviceJobs/deviceJobTags.js
```diff
@@ -1,20 +1,304 @@
...
+
+  componentWillReceiveProps(nextProps) {
+    if (nextProps.devices && (this.props.devices || []).length !== nextProps.devices.length) {
+      this.populateState(nextProps.devices);
+    }
+  }
+
+  populateState(devices) {
+    // Create a stream of devices
+    const deviceStream = Observable.of(devices).flatMap(_ => _);
```
spryor: I think this is a better approach. The current version is roughly n^2. This new version is roughly 2n.

I wrote this blind btw so might need some fixing up.

```
Observable.from(devices)
  .map(({ tags }) => new Set(Object.keys(tags)))
  .reduce((commonTags, deviceTags) => 
    commonTags
      ? new Set([...commonTags].filter(tag => deviceTags.has(tag)))
      : deviceTags
  ) // At this point, a stream of a single event. A common 
  .flatMap(commonTagsSet =>
    Observable.from(devices)
      .flatMap(({ tags }) => Object.entries(tags))
      .filter(([ tag ]) => commonTagsSet.has(tag))
  )
  .distinct(([ tagName, tagVal ]) => `${tagName} ${tagVal}`)
  .reduce((acc, [ tagName, tagVal ]) => update(acc, {
    [tagName]: { $autoArray: { 
      $push: [ tagVal ] 
    }}
  }), {})
  .flatMap(tagToValMap => Object.entries(tagToValMap))
  .reduce(
    (newState, [ name, values ]) => {
      const value = values.length === 1 ? values[0] : TagJobConstants.multipleValues;
      const type = values.every(isNumeric) ? TagJobConstants.numberType : TagJobConstants.stringType;
      return update(newState, {
        commonTags: { $push: [{ name, value, type }] }
      });
    },
    initialState
  ).subscribe(newState => this.setState(newState));
```

src/components/pages/devices/flyouts/deviceJobs/deviceJobTags.js
spryor: Hugh is correct. Since this is in the render function it will be remade when the props change. You only need to wrap the message in a function when you need to grab the props in a function closure.

src/components/pages/devices/flyouts/deviceJobs/deviceJobTags.js
```diff
@@ -1,20 +1,304 @@
...
+  apply = (event) => {
+    event.preventDefault();
+    if (this.formIsValid()) {
+      this.setState({ isPending: true });
+
+      const { devices } = this.props;
+      const { commonTags, deletedTags } = this.state;
+      const request = toSubmitTagsJobRequestModel(devices, this.state);
+
+      this.subscription = IoTHubManagerService.sumbitJob(request)
```
spryor: Typo `sumbitJob`

src/components/pages/devices/flyouts/deviceJobs/deviceJobTags.js
```diff
@@ -1,20 +1,304 @@
...
+        .subscribe(
+          ({ jobId }) => {
+            this.setState({ jobId, successCount: devices.length, isPending: false, changesApplied: true });
+            this.props.updateTags({ deviceId: devices.map(({ id }) => id), commonTags, deletedTags });
+          },
+          errorResponse => {
+            this.setState({ error: errorResponse.errorMessage, isPending: false, changesApplied: true });
```
spryor: For handling the error. You can save it directly and use the `AjaxError` component I made and placed in shared. It has a unified error formatting and handles all the error translation stuff. You can find examples in the dashboard panels

src/components/pages/devices/flyouts/deviceJobs/deviceJobTags.js
```diff
@@ -1,20 +1,304 @@
...
+              }
+              {
+                Object.keys(commonTags).length > 0 &&
+                tagLinks.map(({ name, value, type, edited, error }, idx) => [
+                  <Row key={idx} className={error ? "error-data-row" : ""}>
+                    <Cell className="col-3">
+                      <FormControl className="shorter" type="text" link={name} errorState={!!error} />
+                    </Cell>
+                    <Cell className="col-3">
+                      <FormControl className="shorter" type="text" link={value} errorState={!!error} /></Cell>
```
spryor: Haha, `shorter`. I'm not going to block the PR but put a TODO in the formControl to rename these. We should probably go with something like extra-small, small, regular .. etc. Better yet set it up to use 100% width correctly.

src/components/shared/pcsGrid/pcsGrid.js
```diff
@@ -81,6 +99,9 @@ export class PcsGrid extends Component {
 
   /** When a row is hard selected, try to fire a hard select event, plus any props callbacks */
   onSelectionChanged = () => {
+    const currentHardSelectIds = this.gridApi.getSelectedRows().map(({ id }) => ({ id }));
```
spryor: This isn't actually going to work since this breaks the generality of the PcsGrid. The grid can't have any dependency on the shape of the data.

src/services/models/iotHubManagerModels.js
```diff
@@ -54,6 +55,23 @@ export const toJobsModel = (response = []) => response.map(job => camelCaseResha
   'type': 'type'
 }));
 
+export const toSubmitTagsJobRequestModel = (devices, { jobName, commonTags, deletedTags }) => {
+  const jobId = jobName ? jobName + '-' + uuid() : uuid();
```
spryor: Does the UI add the id or does the service?

src/services/models/iotHubManagerModels.js
```diff
@@ -54,6 +55,23 @@ export const toJobsModel = (response = []) => response.map(job => camelCaseResha
...
+  const deviceIds = devices.map(({ id }) => `'${id}'`).join(',');
+  let Tags = {};
```
spryor: Make this `const`

src/services/models/iotHubManagerModels.js
```diff
@@ -54,6 +55,23 @@ export const toJobsModel = (response = []) => response.map(job => camelCaseResha
...
+  commonTags.map(({ name, value }) => (Tags[name] = value));
```
spryor: Change these `map`s to `forEach`

src/store/reducers/devicesReducer.js
```diff
@@ -82,6 +82,22 @@ const insertDeviceReducer = (state, { payload }) => {
   });
 };
 
+const updateTagsReducer = (state, { payload }) => {
+  let updatedTagData = {};
+  payload.commonTags.map(({ name, value }) => (updatedTagData[name] = value));
```
spryor: forEach

src/store/reducers/devicesReducer.js
```diff
@@ -82,6 +82,22 @@ const insertDeviceReducer = (state, { payload }) => {
   });
 };
 
+const updateTagsReducer = (state, { payload }) => {
+  let updatedTagData = {};
```
spryor: const

src/store/reducers/devicesReducer.js
```diff
@@ -82,6 +82,22 @@ const insertDeviceReducer = (state, { payload }) => {
...
+  payload.commonTags.map(({ name, value }) => (updatedTagData[name] = value));
+
+  const updatedDevices = payload.deviceId
+    .map((id) => state.entities[id])
```
spryor: Use the getEntities selector here.

src/components/pages/devices/flyouts/deviceJobs/deviceJobTags.js
```diff
@@ -1,20 +1,304 @@
...
+
+      const { devices } = this.props;
+      const { commonTags, deletedTags } = this.state;
+      const request = toSubmitTagsJobRequestModel(devices, this.state);
+
+      this.subscription = IoTHubManagerService.sumbitJob(request)
+        .subscribe(
+          ({ jobId }) => {
+            this.setState({ jobId, successCount: devices.length, isPending: false, changesApplied: true });
+            this.props.updateTags({ deviceId: devices.map(({ id }) => id), commonTags, deletedTags });
```
spryor: Rename `deviceId` to `deviceIds` to avoid confusion in the reducer

src/store/reducers/devicesReducer.js
```diff
@@ -82,6 +82,22 @@ const insertDeviceReducer = (state, { payload }) => {
...
+    .map((device) => update(device, { tags: { $merge: updatedTagData } }))
```
spryor: You can do all the updates in one go.

```
const updatedDevices = payload.deviceIds
  .map(id => update(getEntities(state)[id], {
    tags: { $merge: updatedTagData, $unset: payload.deletedTags }
  }));
```

src/store/reducers/devicesReducer.js
mechaffin: Except, again... it gets updated. What am I missing here?

src/store/reducers/devicesReducer.js
mechaffin: I kept looking for an example of how to do this!

src/services/models/iotHubManagerModels.js
spryor: The object gets updated but the reference to the object does not.

src/store/reducers/devicesReducer.js
spryor: See answer above

src/components/shared/pcsGrid/pcsGrid.js
```diff
@@ -62,6 +65,22 @@ export class PcsGrid extends Component {
     }
   }
 
+  /** Ensure that hard selected rows are maintained by their ids, even when the actual data may change */
+  componentDidUpdate(nextProps, nextState) {
+    const { idName = 'id' } = this.props; //Allow override of the name of the identifier property on the data items.
```
spryor: The pcsGrid already has a prop that knows how to access an id called `getSoftSelectId`. If you want to allow this functionality use that.

src/components/shared/pcsGrid/pcsGrid.js
```diff
@@ -81,6 +100,9 @@ export class PcsGrid extends Component {
 
   /** When a row is hard selected, try to fire a hard select event, plus any props callbacks */
   onSelectionChanged = () => {
+    const currentHardSelectIds = this.gridApi.getSelectedRows().map(({ id }) => ({ id }));
```
spryor: This `id` is not dynamic.

src/components/pages/devices/flyouts/deviceJobs/deviceJobTags.js
```diff
@@ -1,20 +1,302 @@
...
+  return object ?
+    update(object, value):
+    update([], value);
+});
+
+const isNumeric = value => typeof value === 'number' || !isNaN(parseInt(value, 10));
+const isAlphaNumericRegex = /^[a-zA-Z0-9]*$/;
+const nonAlphaNumeric = x => !x.match(isAlphaNumericRegex);
+
+const TagJobConstants = {
```
spryor: camelCase

src/store/reducers/devicesReducer.js
```diff
@@ -82,6 +82,21 @@ const insertDeviceReducer = (state, { payload }) => {
   });
 };
 
+const updateTagsReducer = (state, { payload }) => {
+  const updatedTagData = {};
+  payload.commonTags.forEach(({ name, value }) => (updatedTagData[name] = value));
```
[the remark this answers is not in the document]
mechaffin: First, you want me to change map to forEach.  Now, change to reduce. I'm confused as to when to use what. Seems like none of them are "wrong" and they all would work.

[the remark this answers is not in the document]
mechaffin: Nuances of javascript syntax being what they are.  I cannot seem to get reduce to work without also using immutability helper.  Since that seems like overkill and unnecessary overhead in this situation, I am leaving the forEach.

src/components/pages/devices/flyouts/deviceJobs/deviceJobTags.js
mechaffin: camelCase would be inconsistent with previous PRs with similar exterrnal entities... see deviceNew.js, iotHubManagerModels.js

spryor:
Closing since these changes went in another PR

PR #945, opened 2018-04-25
Fixed untracked files - for gitIgnore

isaac73:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

[https://stackoverflow.com/questions/11451535/gitignore-is-not-working](url)
...

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/945)


.env
```diff
@@ -1,5 +0,0 @@
-# Copyright (c) Microsoft. All rights reserved.
```
ppathan: This file is getting deleted, is that expected?

isaac73: It's not expected. Need to dig deeper into this.

PR #946, opened 2018-04-25
Manage device groups flyout services

spryor:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

Add button to open the flyout
Add error and isPending handlers
Add service calls to the manage device group flyout


src/utilities/svgs.js
```diff
@@ -11,13 +11,13 @@ import ContosoIconPath from 'assets/icons/contoso.svg';
 import CopyIconPath from 'assets/icons/copy.svg';
 import CriticalIconPath from 'assets/icons/critical.svg';
 import DeviceChillerIconPath from 'assets/icons/deviceChiller.svg';
-import DeviceGenericIconPath from 'assets/icons/deviceIcon.svg';
 import DeviceElevatorIconPath from 'assets/icons/deviceElevator.svg';
 import DeviceEngineIconPath from 'assets/icons/deviceEngine.svg';
+import DeviceGenericIconPath from 'assets/icons/deviceIcon.svg';
```
spryor: Sorted the order here and added ManageFiltersIcon

src/services/configService.js
```diff
@@ -15,6 +21,29 @@ export class ConfigService {
...
+  }
+
+  /** Delete a device group */
+  static deleteDeviceGroup(id) {
+    return HttpClient.delete(`${ENDPOINT}devicegroups/${id}`)
+      .map(_ => id);
+  }
+
+  /** Returns the account's device group filters */
+  static getDeviceGroupFilters() {
```
ppathan: nit: `getDeviceGroups`

src/services/configService.js
```diff
@@ -15,6 +21,29 @@ export class ConfigService {
...
+    return HttpClient.get(`${ENDPOINT}devicegroupfilters`)
+      .map(toDeviceGroupFiltersModel);
```
ppathan: nit: `toDeviceGroupsModel`

src/store/reducers/appReducer.js
```diff
@@ -104,14 +111,16 @@ const deviceGroupListSchema = new schema.Array(deviceGroupSchema);
...
   ...errorPendingInitialState,
   deviceGroups: {},
+  deviceGroupFilters: {},
   activeDeviceGroupId: undefined,
   theme: 'dark',
   version: undefined,
   releaseNotesUrl: undefined,
   logo: svgs.contoso,
-  name: 'companyName',
+  name: 'Company',
```
ppathan: This is coming from translations.json which is `companyName` won't this break?

src/services/configService.js
spryor: This is not loading the device groups. It is loading the filters used to make device groups.

src/services/configService.js
spryor: See above

src/store/reducers/appReducer.js
spryor: oops, good catch

src/services/configService.js
ppathan: Ahh...get it now

PR #950, opened 2018-04-26
Pass the theme to the tsiCharts

spryor:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

Make the tsiCharts theme change with the app theme


src/components/pages/dashboard/dashboard.js
```diff
@@ -385,6 +387,7 @@ export class Dashboard extends Component {
               telemetry={telemetry}
               isPending={telemetryIsPending}
               error={deviceGroupError || telemetryError}
+              theme={theme}
               colors={chartColorObjects}
```
mechaffin: Shouldn't the colors come from the theme?

src/components/pages/maintenance/maintenance.js
```diff
@@ -150,7 +150,7 @@ export class Maintenance extends Component {
   onTimeIntervalChange = (timeInterval) => this.setState({ timeInterval }, () => this.getData());
 
   render() {
-    const { rulesEntities, deviceEntities, rulesIsPending, t, history } = this.props;
+    const { rulesEntities, deviceEntities, rulesIsPending, theme, t, history } = this.props;
```
mechaffin: At what point should these be on separate lines? I recall being asked to make separate lines when lists get long. I've been using 3 as a rule of thumb.

src/components/pages/dashboard/dashboard.js
spryor: They should but that is a larger item than this task. Currently, if you switch to the light theme, the tsi charts stay on the default dark theme where the legends and axis are white. So you can't see them, This PR just makes the charts readable when the light theme is turned on.

src/components/pages/maintenance/maintenance.js
spryor: This rule isn't strictly enforced but in the past we have been using character count as the cut off. I think the cutoff is 90 something chars. It is a judgment call though and I'm fine with this (currently 86 chars long anyway).

PR #978, opened 2018-05-02
Fix undefined hostname on device create

molly-moen:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Hostname was undefined in connection string due to casing error.

**Checklist:**

- [x] All tests passed


src/services/models/iotHubManagerModels.js
```diff
@@ -24,7 +24,7 @@ export const toDeviceModel = (device = {}) => {
     'c2DMessageCount': 'c2DMessageCount',
     'enabled': 'enabled',
     'lastStatusUpdated': 'lastStatusUpdated',
-    'iotHubHostName': 'iotHubHostName',
+    'ioTHubHostName': 'ioTHubHostName',
```
ppathan: Isn't this naming convention wrong? iotHubHostName is camel case no idea what ioTHubHostName is :)

molly-moen: ioTHubHostName is what our code will translate as "camel case" from IoTHubHostName, which is what is returned. We can map ioTHubHostName to iotHubHostName in this and use that as the variable name

ppathan: Yeah can we do that keep client side iotHubHostName?

src/services/models/iotHubManagerModels.js
```diff
@@ -24,7 +24,7 @@ export const toDeviceModel = (device = {}) => {
     'c2DMessageCount': 'c2DMessageCount',
     'enabled': 'enabled',
     'lastStatusUpdated': 'lastStatusUpdated',
-    'ioTHubHostName': 'ioTHubHostName',
+    'ioTHubHostName': 'iotHubHostName',
```
dluc: seems strange that the 't' is capital case on the left, lower case on the right. Could we use the same format?

molly-moen: Here is what is happening:
From IotHubManager we get IoTHubHostName
The UI auto-camelcases everything, which turns into ioTHubHostName
we want it to be iotHubHostName in the code, so this maps ioTHubHostName coming from our request to iotHubHostName, what we want the variable to be called.

dluc: thanks! that would be a great comment to add future readers

PR #985, opened 2018-05-02
Show a message over the analytics panel if no data is available

spryor:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Currently, if the analytics are empty, either nothing or the previous data will be shown. This change will hide the old data.


src/components/pages/dashboard/panels/analytics/analyticsPanel.js
```diff
@@ -35,7 +36,7 @@ export class AnalyticsPanel extends Component {
   handleWindowFocus = () => this.setState({ renderChart: true });
 
   componentDidMount() {
-    this.barChart = new this.tsiClient.ux.BarChart(document.getElementById(barChartId));
+    this.barChart = new this.tsiClient.ux.BarChart(document.getElementById(barChartId+'wrong'));
```
avanikp: nit: Add space before and after +

src/components/pages/dashboard/panels/analytics/analyticsPanel.js
```diff
@@ -121,6 +122,10 @@ export class AnalyticsPanel extends Component {
               }
             </div>
           </div>
+          {
+            (!showOverlay && !topAlerts.length && !Object.keys(alertsPerDeviceId).length)
```
avanikp: can topAlerts be undefined ? Need a check for that ?
Same for Object.keys(alertsPerDeviceId)

src/components/pages/dashboard/panels/analytics/analyticsPanel.js
ppathan: Why is this `wrong` hardcoded?

src/components/pages/dashboard/panels/analytics/analyticsPanel.js
spryor: No, they are always defined in the dashboard.

PR #988, opened 2018-05-03
Retry on hourly 401 auth error codes for dashboard & analytics

jillcary:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

The ADAL js library sets the bearer token to null once an hour when it attempts to renew an expiring token. This causes a 401 not logged in response code from the back-end briefly until ADAL js gets the new token. 

This error was not categorized as "retryable" in the front-end and it would persist in the redux store. This was causing a "User is not logged in" error message on the dashboard and rules page that would persist, even though the api calls to the back-end were succeeding. 

This is a screenshot of the 401 causing the error in the UI, notice the successful requests after the errors.
image

**Checklist:**

- [x] All tests passed


src/app.config.js
```diff
@@ -15,7 +15,7 @@ const Config = {
   defaultAjaxTimeout: 20000, // 20s
   maxRetryAttempts: 2,
   retryWaitTime: 2000, // On retryable error, retry after 2s
-  retryableStatusCodes: new Set([ 0, 502, 503 ]),
+  retryableStatusCodes: new Set([ 0, 502, 503, 401 ]),
```
zhang-hua: put 401 before 502?

src/components/pages/dashboard/dashboard.js
```diff
@@ -81,7 +83,11 @@ export class Dashboard extends Component {
           .flatMap(_ => TelemetryService.getTelemetryByDeviceIdP1M(deviceIds))
       )
       .flatMap(transformTelemetryResponse(() => this.state.telemetry))
-      .map(telemetry => ({ telemetry, telemetryIsPending: false })); // Stream emits new state
+      .map(telemetry => ({ telemetry, telemetryIsPending: false })) // Stream emits new state
+      // Classify errors as retryable or not
+      .catch(error => Observable.throw(classifyError(error)))
```
spryor: I don't think you need this line. If there is an error at this point, the error will already have been classified in the HttpClient. I think classifying it here might actually break the error logic in the panels (by changing the structure of the error message).

PR #991, opened 2018-05-03
Have simulation toggle apply automatically

molly-moen:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

Currently simulation toggle needs to be toggled, then apply button must be clicked in order to turn on/off simulation. This change makes call to simulation service as soon as toggle is clicked.

**Checklist:**

- [x] All tests passed

### DIFF
--- a/src/components/app/flyouts/settings/settings.js
+++ b/src/components/app/flyouts/settings/settings.js
@@ -22,7 +22,7 @@ export class Settings extends LinkedComponent {
       applicationName: '',
       loading: false,
       toggledSimulation: false,
-      madeLogoUpdate: false,
+      madeLogoUpdate: false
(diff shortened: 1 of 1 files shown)


src/components/app/flyouts/settings/settings.js
```diff
@@ -75,8 +62,18 @@ export class Settings extends LinkedComponent {
     this.setState({ [name]: value });
   };
 
+  onSimulationChange = ({ target }) => {
+    this.onChange({target});
```
spryor: Since you are setting the state in this method already, it would be better to set the state for the onChange here as well instead of having multiple back to back state updates.

PR #999, opened 2018-05-08
Better parsing for exceptions

mechaffin:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Implement better parsing based on IotHub exceptions. They are different coming from dotnet and java. They also include ugly/scary looking text. So we want to parse them looking for known "error codes" that we can then give the user a more friendly message.

**Checklist:**

- [x] All tests passed


src/services/models/ajaxModels.js
```diff
@@ -11,10 +12,34 @@ export class AjaxError {
...
+    this.errorMessage = resp.InnerExceptionMessage || resp.ExceptionMessage || resp.Message || resp.Error || ajaxError.message || `An unknown error occurred`;
     this.status = ajaxError.status;
+
     // Log all errors in the console
     console.error(ajaxError);
+
+    // For general application errors, figure out a more detailed message if possible.
+    // In dotNet, the ExceptionMessage may contain JSON that can be further parsed.
+    if (this.status >= 500 && this.errorMessage) {
+      this[parseApplicationMessage](((JSON.parse(this.errorMessage) || {}).Message) || this.errorMessage);
```
spryor: I don't see the value of using the symbol here. Just seems to make things more confusing.

public/locales/en/translations.json
```diff
@@ -16,6 +16,7 @@
   },
   "errorFormat": "Error: {{message}}",
   "errorCode": {
+    "maxJobCountExceeded": "We were not able to create the job because the maximum job count has been exceeded.",
```
spryor: I don't think we should have this in our translations since doing so opens the door to a world of problems. We now take the possible responsibility for storing translations strings for not just our own but error strings/status codes used by  other teams. Of course we then take a dependency on their error code.

src/services/models/ajaxModels.js
```diff
@@ -29,10 +54,14 @@ export class AjaxError {
       return 'errorCode.notAuthorized';
     } else if (this.status === 404) { // Endpoint not found
       return 'errorCode.notFound';
+    } else if (this.status === 429) { // Max Active Job Count Exceeded
```
spryor: I don't think we can necessarily guarantee this. Might it be possible for one of our micro services to return a 429? If so we'll show the wrong error message. We don't have the right amount of granularity here to avoid those problems.

mechaffin: This one isn't supposed to be there.  Sorry... I didn't undo this part.

public/locales/en/translations.json
mechaffin: I hear you.  If we don't translate ourselves, then ( 1 ) messages will contain ugly/scary looking information, and ( 2 ) messages from dotNet and java are different. 
Also, do we have a way to tell the server what language to give us messages in?  Because if we don't, then there's a high probability of language mismatches.

src/services/models/ajaxModels.js
mechaffin: That seemed like the way to make it private... but if we aren't going to translate... then its not going to be needed anyway.

public/locales/en/translations.json
mechaffin: Here's what the dotnet message looks like now... without adding our own translation....  image

PR #1005, opened 2018-05-10
Move Azure map initialization to a seperate component

spryor:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

The current implementation of azure maps is giving console errors related to calling azure map apis before the map is completely initialized. This PR wraps the azure map in it's own component that will provide the map reference only after the initialization is complete. This prevents making calls to the map api before it is ready, thus removing the console errors.


src/components/pages/dashboard/panels/map/azureMap.js
```diff
@@ -0,0 +1,48 @@
...
+      if (isFunc(this.props.onMapReady)) {
+        this.props.onMapReady(this.map);
+      }
+    });
+  }
+
+  render() {
+    return <div id="map"></div>;
+  }
+
```
mechaffin: extra line

src/components/pages/dashboard/panels/map/mapPanel.js
```diff
@@ -33,62 +34,33 @@ const deviceToMapPin = ({ id, properties, type }) =>
...
-  initializeMap(azureMapsKey) {
-    const center = new AzureMaps.data.Position(...Config.mapCenterPosition);
-    this.map = new AzureMaps.Map('map', {
-      'subscription-key': azureMapsKey,
-      center,
-      zoom: 11
-    });
-
+  onMapReady = (map) => {
+    this.map = map;
```
mechaffin: Set this.map at the end of this function, just before the call to calculatePins.  That way there's no chance that map is there but the layers aren't when calculatePins is called from other places (like componentWillReceiveProps).

src/components/pages/dashboard/panels/map/azureMap.js
spryor: I added the newline intentionally.

src/components/pages/dashboard/panels/map/mapPanel.js
spryor: Nice catch!

PR #1007, opened 2018-05-10
Never update the azureMap component

spryor:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Never update the azureMap component

### DIFF
--- a/src/components/pages/dashboard/panels/map/azureMap.js
+++ b/src/components/pages/dashboard/panels/map/azureMap.js
@@ -26,6 +26,11 @@ export class AzureMap extends Component {
     if (this.map) this.map.remove();
   }
 
+  shouldComponentUpdate(nextProps) {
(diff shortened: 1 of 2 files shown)


src/components/pages/dashboard/panels/map/azureMap.js
```diff
@@ -26,6 +26,11 @@ export class AzureMap extends Component {
     if (this.map) this.map.remove();
   }
 
+  shouldComponentUpdate(nextProps) {
+    // Component props never result in a dom updates from React
+    return false;
```
mechaffin: If it will never update, even when/if the azureMapsKey changes, then why would the mapPanel say shouldComponentUpdate is true when the azureMapsKey changes?  If the azureMap won't respect the change, then its caller shouldn't care either, right?

PR #1013, opened 2018-05-15
Add question icon on the header

spryor:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

Add question icon on the header to open docs


src/components/app/header/header.js
```diff
@@ -41,18 +64,45 @@ class Header extends Component {
...
-          </button>
-          {
-            this.state.dropdownExpanded &&
-            <div className="profile-dropdown">
-              <button className="dropdown-item" onClick={this.logout}>{ this.props.t('header.logout') }</button>
-            </div>
-          }
+          <div className="menu-container">
+            <button className="item-icon profile menu-trigger" onClick={this.toggleDropdown(profileDropdown)}>
+              <img src={ProfileImagePath} alt={ this.props.t('header.logout') } />
```
avanikp: This should be text for "Profile"

src/components/app/header/header.js
```diff
@@ -9,29 +10,66 @@ import ProfileImagePath from 'assets/images/profile.png';
...
+    translationId: 'header.getStarted',
+    url: 'https://docs.microsoft.com/en-us/azure/iot-suite/iot-suite-remote-monitoring-monitor'
+  },
+  {
+    translationId: 'header.documentation',
+    url: 'https://docs.microsoft.com/en-us/azure/iot-suite'
+  },
+  {
+    translationId: 'header.sendSuggestion',
+    url: 'https://feedback.azure.com/forums/321918-azure-iot'
```
ppathan: Is there a specific link for azure iot solutions?

mechaffin: Should these URLs be in config so a "code" change isn't needed?

src/components/app/header/header.js
```diff
@@ -41,18 +79,40 @@ class Header extends Component {
         </div>
         <div className="label">{ this.props.t('header.appName') }</div>
         <div className="items-container">
+          <div className="menu-container">
```
mechaffin: We have other places that will have menus, right? For example, where there will be a menu of actions for device alerts... which currently is just a Link to the alert, but that they want to be a menu containing the link as well as actions to ack or close right there without navigating. I think there are other similar "in grid" menus as well.
Any reason to not go ahead and create a menu component now?

src/components/app/header/header.js
spryor: I don't know. These are the links from the spec that I got.

src/utilities/svgs.js
```diff
@@ -33,14 +33,15 @@ import LoadingToggleIconPath from 'assets/icons/loadingToggle.svg';
...
+import QuestionMarkIconPath from 'assets/icons/questionMark.svg';
 import RadioSelectedIconPath from 'assets/icons/radioSelected.svg';
 import RadioUnselectedIconPath from 'assets/icons/radioUnselected.svg';
 import ReconfigureIconPath from 'assets/icons/reconfigure.svg';
 import RefreshIconPath from 'assets/icons/refresh.svg';
 import RuleDisabledIconPath from 'assets/icons/ruleDisabled.svg';
 import RuleEnabledIconPath from 'assets/icons/ruleEnabled.svg';
-import SettingsIconPath from 'assets/icons/settings.svg';
 import SIMManagementIconPath from 'assets/icons/SIMManagement.svg';
+import SettingsIconPath from 'assets/icons/settings.svg';
```
mechaffin: Alphabetically... se comes before si... so the old location would be correct.

src/components/app/header/header.js
```diff
@@ -9,29 +10,66 @@ import ProfileImagePath from 'assets/images/profile.png';
 
 import './header.css';
 
+const docsDropdown = 'docsDropdown';
+const profileDropdown = 'profileDropdown';
+
+const parentHasClass = (element, ...searchClasses) => {
```
mechaffin: This is sort of weird. If menu functionality was implemented and encapsulated in a component (or set of them), then this shouldn't be needed.

src/utilities/svgs.js
spryor: This is probably coming from how VS code line sort is handling upper vs lowercase. Best to leave it with the default behavior or else it will be undone later on anyway.

src/components/app/header/header.js
spryor: In the future we will have menus but I'm not building one right now for this tasks. Rather, I'm reusing what we already have. This may well be a starting point for a menu in the future. To build a full on reusable menu would take more thinking into how to shared common event listeners, etc.

src/components/app/header/header.js
spryor: True, but I'm building a reusable menu atm. See above comment.

src/components/app/header/header.js
```diff
@@ -71,21 +86,16 @@ class Header extends Component {
...
-                  to="https://docs.microsoft.com/en-us/azure/iot-suite/">
-                  { this.props.t('header.documentation') }
-                </Link>
-                <Link className="menu-item"
-                  target="_blank"
-                  to="https://feedback.azure.com/forums/321918-azure-iot ">
-                  { this.props.t('header.sendSuggestion') }
-                </Link>
+                {
+                  docLinks.map(({ url, translationId }) =>
```
avanikp: nice refactor!

PR #1014, opened 2018-05-15
ON HOLD- UI change related to cache change in APIs

isaac73:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Service changes due to cache logic change in Endpoints
...

**Checklist:**

- [ ] All tests passed


src/services/iotHubManagerService.js
```diff
@@ -46,4 +48,12 @@ export class IoTHubManagerService {
     return HttpClient.delete(`${ENDPOINT}devices/${id}`)
       .map(() => ({ deletedDeviceId: id }));
   }
+
+    /** Returns the account's device group filters */
+    static getDeviceProperties() {
+      return Observable.forkJoin(
+      HttpClient.get(`${ENDPOINT}deviceProperties`).map(toDevicePropertiesModel),
```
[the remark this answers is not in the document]
dluc: rather than hardcoding, we should have methods to return these URLs, in a reusable fashion

1. `${ENDPOINT}deviceProperties`
2. `${Config.serviceUrls.deviceSimulation}deviceProperties`

Since it's a cross-cutting concern across the entire codebase, I'd suggest to discuss this with @spryor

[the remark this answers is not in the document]
dluc: it's also odd that in the first case the app uses a constant (`ENDPOINT = Config.serviceUrls.iotHubManager`) but in the second it references the config directly (`${Config.serviceUrls.deviceSimulation}`)

spryor: Did you test this code? `toDevicePropertiesModel` model returns an object but you are using the spread operator into an array, which doesn't make sense and seems like it should cause a type error. The ajax calls don't return arrays so you want to merge the internal objects `{ reported, tags }`. I think this should do what you want (note `import update from 'immutability-helper';`):
```
return Observable.merge(
  HttpClient.get(`${ENDPOINT}deviceProperties`)
  HttpClient.get(`${Config.serviceUrls.deviceSimulation}deviceProperties`)
)
.map(toDevicePropertiesModel)
.reduce(
  (acc, props) => update(acc, {
    reported: { $push: props.reported },
    tags: { $push: props.tags }
  }), 
  { reported: [], tags: [] }
);
```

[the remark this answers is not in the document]
spryor: @dluc Why do you think these should be wrapped in functions? The services classes serve as a single source of truth for service calls so urls (other than the configurable endpoints) should never change or be referenced directly outside these files. What is the benefit of adding an extra layer of complexity with functions?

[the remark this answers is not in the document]
dluc: within the classes, I would define either a property or a method for each collection. For example, IoT Hub manager provides devices, deviceproperties, queries and jobs, so I'd consider adding something like:

devicesEndpoint = ${ENDPOINT}devices
devicePropertiesEndpoint = ${ENDPOINT}deviceProperties
queriesEndpoint = ${ENDPOINT}deviceProperties
jobsEndpoint = ${ENDPOINT}jobs

A separate observation, in the backend we code service classes by collection, while in the UI (if understand correctly) there is a service class per microservice. Have you considered having a "devices" class, a "deviceProperties" class, a "jobs" class etc?  In that case, each class would deal with a single endpoint.

(btw, these are not blocking considerations, just an observation about the project structure)

src/services/models/iotHubManagerModels.js
```diff
@@ -167,3 +167,16 @@ export const toNewDeviceRequestModel = ({
...
 }
+
+export const toDevicePropertiesModel = (iotResponse, dsResponse) => {
+  const iotModel = reshape(iotResponse, {
+    'Reported': 'reported',
+    'Tags': 'tags'
+  });
+  const dsReported = getItems(dsResponse).map(({ Id }) => Id);
+  const reportedPropertiesUnion = new Set([...iotModel.reported, ...dsReported])
+  return {
```
spryor: Taking a second look we can probably shrink this a bit more.

```
export const toDevicePropertiesModel = (iotResponse, dsResponse) => {
  const dsReported = getItems(dsResponse).map(({ Id }) => Id);
  const reportedPropertiesUnion = new Set([...iotResponse.Reported, ...dsReported]);
  return {
    tags: iotResponse.Tags,
    reported: [...reportedPropertiesUnion]
  };
}
```

src/services/iotHubManagerService.js
```diff
@@ -46,4 +48,14 @@ export class IoTHubManagerService {
...
       .map(() => ({ deletedDeviceId: id }));
   }
+
+  /** Returns the account's device group filters */
+  static getDeviceProperties() {
+    return Observable
+      .forkJoin(
+        HttpClient.get(`${ENDPOINT}deviceproperties`),
+        HttpClient.get(`${Config.serviceUrls.deviceSimulation}devicemodelproperties`)
+    )
```
spryor: Indentation is off here.

src/services/models/iotHubManagerModels.js
isaac73: done

src/services/iotHubManagerService.js
isaac73: done

src/services/iotHubManagerService.js
```diff
@@ -46,4 +48,14 @@ export class IoTHubManagerService {
     return HttpClient.delete(`${ENDPOINT}devices/${id}`)
       .map(() => ({ deletedDeviceId: id }));
   }
+
+  /** Returns the account's device group filters */
+  static getDeviceProperties() {
+    return Observable
+      .forkJoin(
+        HttpClient.get(`${ENDPOINT}deviceproperties`),
+        HttpClient.get(`${Config.serviceUrls.deviceSimulation}devicemodelproperties`)
```
dluc: the correct URL is camelCase: "deviceModelProperties"

src/services/models/iotHubManagerModels.js
```diff
@@ -167,3 +167,12 @@ export const toNewDeviceRequestModel = ({
         }
   };
 }
+
+export const toDevicePropertiesModel = (iotResponse, dsResponse) => {
+  const dsReported = getItems(dsResponse).map(({ Id }) => Id);
+  const reportedPropertiesUnion = new Set([...iotResponse.Reported, ...dsReported])
+  return {
+      tags: iotResponse.Tags,
```
dluc: the response from the web service should be a list of "Items" like all the other microservices

src/services/iotHubManagerService.js
isaac73: I've refereed few other URLs in our services, we don't use camel case. Besides it shouldn't matter.

src/services/models/iotHubManagerModels.js
```diff
@@ -167,3 +167,7 @@ export const toNewDeviceRequestModel = ({
         }
   };
 }
+
+export const toDevicePropertiesModel = (iotResponse, dsResponse) => {
+  return [...(new Set([...getItems(iotResponse), ...getItems(dsResponse)]))];
```
spryor: This will work but as mentioned it would probably be helpful to break this down into an extra variable or two for readability purposes. Maybe break the set into a separate variable before converting it into an array.

src/services/iotHubManagerService.js
isaac73: @dluc camelcase doesen't work with Java Play  framework

PR #1016, opened 2018-05-16
Add alert ack indicator

mechaffin:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

Add alert ack indicator.  Usually, it happens so fast that you won't see it at all.  But if there is a delay, it'll be there.

**Checklist:**

- [x] All tests passed

### DIFF
--- a/src/components/pages/maintenance/ruleDetails/ruleDetails.js
+++ b/src/components/pages/maintenance/ruleDetails/ruleDetails.js
@@ -6,7 +6,7 @@ import { Observable, Subject } from 'rxjs';
 
 import Config from 'app.config';
 import { RulesGrid } from 'components/pages/rules/rulesGrid';
-import { AjaxError, Btn, PageContent, ContextMenu, RefreshBar } from 'components/shared';
+import { AjaxError, Btn, PageContent, ContextMenu, RefreshBar, Indicator } from 'components/shared';
 import { svgs, joinClasses, renderUndefined } from 'utilities';
 import { DevicesGrid } from 'components/pages/devices/devicesGrid';
 import { TelemetryChart, transformTelemetryResponse, chartColorObjects } from 'components/pages/dashboard/panels/telemetry';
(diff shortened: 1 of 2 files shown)


src/components/pages/maintenance/ruleDetails/ruleDetails.js
```diff
@@ -132,17 +134,24 @@ export class RuleDetails extends Component {
...
+      isAlertStatusPending: true,
+      alertContextBtns: this.getAlertContextBtns(selectedAlerts, true)
+    });
+
     this.subscriptions.push(
       Observable.from(selectedAlerts)
         .flatMap(({ id }) => TelemetryService.updateAlertStatus(id, status))
         .toArray() // Use toArray to wait for all calls to succeed
         .subscribe(() => {
+          this.setState({ isAlertStatusPending: false });
```
spryor: Set the pending in the `onComplete`. This subscription doesn't have error handling yet so if an error occurs the pending flag will never be unset.

src/components/pages/maintenance/ruleDetails/ruleDetails.js
```diff
@@ -132,17 +134,24 @@ export class RuleDetails extends Component {
     this.subscriptions.forEach(sub => sub.unsubscribe());
   }
 
-  // TODO: Handle error and isPending states
-  updateAlertStatus = (selectedAlerts, status) =>
+  // TODO: Handle error states
+  updateAlertStatus = (selectedAlerts, status) => {
+    this.setState({
+      isAlertStatusPending: true,
+      alertContextBtns: this.getAlertContextBtns(selectedAlerts, true)
```
spryor: Why are you doing this? It should be as simple as setting the flag and checking it to render the indicator in the `render` method inside the `ContextMenu` component.

mechaffin: You'd think that, right?  But when I did it that way... the indicator never showed up... even when I added a 5 second delay.

PR #1020, opened 2018-05-16
Change type color on white theme

Abuen:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

After reviewing the code with the new light theme we assessed that the contrast was so pronounced it appeared "washed out" and we were concerned about eye strain for users.

**Checklist:**

- [x] All tests passed


src/styles/_themes.scss
```diff
@@ -152,52 +152,52 @@ $themes: (
     colorError: $colorError,
 
     // App - START
-    colorSelectionText: #212121,
+    colorSelectionText: #333333,
```
spryor: Whenever you have a color hex value in css that is a repetition of the same value, you can truncate it to a three character shorthand. e.g. `#333333` becomes `#333`. We always use the shorthand when possible ([see the code style guide](https://github.com/Azure/pcs-remote-monitoring-webui/wiki/Code-Guide#css)).

PR #1025, opened 2018-05-31
Add device and rule grid filter inputs

spryor:
# Type of change? 

- [x] Bug fix
- [x] New feature
- [ ] Enhancement

...also fix incorrect rem-fallback definitions


src/components/pages/devices/flyouts/deviceDetails/propertyGrid/propertyGrid.scss
```diff
@@ -42,7 +42,7 @@ $scrollbarWidth: 4px;
   .row {
     display: flex;
     flex-direction: row;
-    @include rem-fallback(padding, 16px 0px 16px 8px);
+    @include rem-fallback(padding, 16px, 0px, 16px, 8px);
```
mechaffin: Is this just a style thing? Because they work without commas and we've been doing them that way all along.  

Also, these changes are not actually related to the stated purpose of the PR... so should be held for another PR focused on cleanup type actions, if they even should be done at all.

src/components/shared/contextMenu/searchInput/searchInput.js
```diff
@@ -0,0 +1,11 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+import React from 'react';
+
+import { joinClasses } from 'utilities';
+
+import './searchInput.css';
+
+export const SearchInput = ({ children, className, ...rest }) => (
+  <input {...rest} type="text" className={joinClasses('context-menu-search-input', className)} />
```
mechaffin: For consistent styling, should this use FormControl instead?

spryor: No, we want different styling here and we don't want the extra dom elements the form control can render (like elements for errors, etcs).

src/components/pages/devices/flyouts/deviceDetails/propertyGrid/propertyGrid.scss
spryor: Nope. The current code is actually in error. It appears to work because the the fallback part of the rem-fallback mixin is preventing the error. 

`@include rem-fallback(margin, 10px 0px);`

will generate 

```
margin: 10px 0px;
margin: 10px 0px/16pxrem; <- the browser won't know what to do with this so it will default to the previous line, thus making it look like the code worked
```

while using the a comma separated parameter list will compile to the correc

```
margin: 10px 0px;
margin: 0.625rem 0rem;
```

PR #1034, opened 2018-06-13
Add time interval to alerts and jobs maintenance pages

mechaffin:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

- Add the device groups drop down and manage device groups button to the rule/alert maintenance page.
- Add the time interval drop down to the rule/alert maintenance page.
- Add the time interval drop down to the jobs maintenance page.

This enables the user to see what time interval is selected and thus what all the counts and such are for.  If the alarm or job is outside the currently selected timeframe, the user can select a different timeframe.  This is not a perfect fix. However, without a significant redesign and/or refactoring of the maintenance pages, this is the best we can do.  Its still far better than the current state.

**Checklist:**

- [x] All tests passed


src/components/pages/maintenance/ruleDetails/ruleDetails.js
```diff
@@ -65,19 +68,23 @@ export class RuleDetails extends Component {
...
-            .merge(
-              this.telemetryRefresh$ // Previous request complete
-                .delay(Config.telemetryRefreshInterval) // Wait to refresh
-                .do(onPendingStart)
-                .flatMap(_ => TelemetryService.getTelemetryByDeviceIdP1M(deviceIds))
-            )
-            .flatMap(transformTelemetryResponse(() => this.state.telemetry))
-            .map(telemetry => ({ telemetry, telemetryIsPending: false }))
+        .switchMap(deviceIds => {
+          if (deviceIds[0]) {
```
spryor: Why check the initial device id instead of checking the length?

mechaffin: The actual value is [""].  So... The array length is 1.  The length of the first item in the array is 0.

mechaffin: I went with the most compact javascript syntax that got the job done... which makes it about as readable as other compact javascript syntax.  Hahaha.

if (deviceIds[0]) {

is much more compact than

if (deviceIds.length === 1 && deviceIds[0].length === 0) {

PR #1035, opened 2018-06-14
Update TsiClient version

spryor:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

Update TsiClient version


ppathan:
Truly a nit on description. But since we take updates to tsi client fairly regularly, i would suggest add the some features that were added in each version that we are trying to take advantage of.

Issue #1033, opened 2018-06-13
The webui crashs when the twin has a nested tags in reported


franji:
# Type of issue 

- [x ] Bug
- [ ] New feature
- [ ] Enhancement

# Description 
The webui crashs when the twin has a nested tags in reported

# Steps to reproduce 
If the twin has nested tags in the reported 
Example:
"reported": {
   "telemetryConfig": {
"sendFrequency": "24h"
   }
}

If we enter the device section and click on some devices with nested tags, the portal crashes.

# Expected behavior 
The webui show the flayout with details about this device.

# Current behavior 
The webui crashs.

# Known workarounds 
...

# Possible solution 
...

# Context and Environment
* Operating System: Windows 10
* GitHub branch/tag: ...

# Screenshot

ppathan:
@isaac73 Can you look at this? It might be related to the work that you are doing with cache logic change and might be already fixed in your PR.

isaac73:
@mechaffin looks like an issue in the dievices grid/flyout. Please take a look.

PR #1036, opened 2018-06-15
Adding toString back on property values so it doesn't blow up on nest…

mechaffin:
#1033 Adding toString() back on property values so it doesn't blow up on nested values.
 

# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

**Checklist:**

- [x] All tests passed


isaac73:
Is this fix tested for a device with nested properties?

mechaffin:
Yes, of course, it is tested.
In order to do so, I removed the code from the device model that removes the Telemetry property (which contains nested info) from the reported properties. With that back in the property listing, the page does indeed blow up. 
Then, made the change... and it didn't blow up and displayed "[object Object]" as expected.  
I reverted the change to the model since that's not part of the fix and was just used so we'd have data to verify with.

Note: If anyone needs to display more/better info for a nested property, they'll have to enhance the code (perhaps using something like a "cell renderer" approach like the grids). That is not part of the base product at this time.

PR #1038, opened 2018-06-22
Remove camelcase object package since it doesn't handle arrays correctly

spryor:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

The package camelcase-object doesn't handle camel casing arrays with string primitives correctly (it will convert the string into an object of characters).  In this PR, I replace that library with a custom implementation that doesn't have that problem.


src/utilities/camelcase.js
```diff
@@ -0,0 +1,75 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+// This code is a copy of the code made available by https://www.npmjs.com/package/camelcase
+// Since the original module isn't supported by create-react-app we are using the code manually here
```
ppathan: Can you explain this sentence? Are we removing the camelcase-Object and using camelcase? I don't understand what create-react-app has to do with this

spryor: We are removing the camelcase-object package but using the camelcase package code which which was being used by camelcase-object. Because of the way camelcase bundling is configured, it doesn't support the camelcase package since it was packaged to be used by node, not the browser. This is a workaround to use the code and circumvent the packaging issue.

PR #1053, opened 2018-08-01
Walkthroughs: Add a New Grid, Add a New Service

mechaffin:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

- Document how to add a new grid to the application (with sample code).
- Document how to add a new service to the application (with sample code).
- Some updates to the walk through for adding a new page.

**Checklist:**

- [x] All tests passed
- [x] The change requires a change to the documentation
- [x] I have updated the documentation accordingly

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/1053)


src/components/pages/_gridExample/gridExample.js
```diff
@@ -0,0 +1,45 @@
+// Copyright (c) Microsoft. All rights reserved.
```
ppathan: Why are we adding _ in the front of the file name. I'm not sure if we should do this to just differentiate whether this is a sample code.

mechaffin: Yes, it was an effort to differentiate. So its easy to find sample code and delete it if you want to fork from our branch and don't want the extra stuff there. 

As we add more walkthroughs and more sample code to go with them, I wanted to be able to give the files good names without having to always use example or sample at the beginning... and if the thing that makes it look like an example isn't at the beginning, then its harder to find to remove if desired.

I felt like it was valuable to have the samples in the code so they could be debugged and understood that way... without having to recreate using the steps in the walkthroughs. It also helps verify that the walkthrough steps actually work and are complete.

I'm open to other ideas.

mechaffin:
Closing this request as it hasn't had a substantive review in over a week. By now, I have a lot more work done on top of this. I'll create a new PR when that work is ready for review. Its difficult to do small PRs for single tasks when it takes so long (and multiple reminders) to get a review.

PR #1062, opened 2018-08-14
Dev Walkthru: add a new Panel to the Dashboard

mechaffin:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

- Document how to add a new Panel to the dashboard page (with sample code).
- Add a WalkthroughExampleRoute so the routes for sample pages will redirect to root if the samples are turned off.

**Checklist:**

- [x] All tests passed
- [x] The change requires a change to the documentation
- [x] I have updated the documentation accordingly

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/1050)


docs/walkthrough/addNewDashboardPanel.md
```diff
@@ -0,0 +1,78 @@
+Walkthrough: Adding a New Panel to the Dashboard
+================================================
+
+The following is for creating a new panel called "**examplePanel**."
```
elvinm: Do you think it might be handy to explain what a panel is in the first place? Perhaps for someone looking to customize this might already be obvious though. What are your thoughts?

src/components/app/app.js
```diff
@@ -44,6 +44,23 @@ if (Config.showWalkthroughExamples) {
   tabConfigs.push(gridExampleTab);
 }
```
elvinm: Can things like this be moved below or to a separate file? Not a huge deal but I think example stuff shouldn't be the first things someone sees. In some instances like const flyoutExampleTab being near all the other tab configurations makes sense. But individual classes like this seems like they can be moved.

docs/walkthrough/addNewDashboardPanel.md
mechaffin: Its a dashboard panel.  I guess I could explicitly state that... but since the title says it and the code is all under the dashboard folder... it seemed redundant.

src/components/pages/dashboard/panels/_examplePanel/examplePanel.scss
```diff
@@ -0,0 +1,14 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+@import 'src/styles/mixins';
+@import 'src/styles/themes';
+
+.example-panel-container {
```
elvinm: The panel-container class already has these properties by default. Should we include both classes in the example panel so that the css isn't needed here? Or is the idea to have a guide on how to customize the display more?

mechaffin: When they add real stuff to their panel, they'll need a place to put it.  That's really why I created this file.

src/components/app/app.js
mechaffin: The final location of the docs and sample code is still to be determined.  For example, the doc folks want to put these with the rest of the docs (i.e. not in the code).  Having working sample code is valuable though...and the docs folks can't do that. So the sample code might move to its own repo... or get removed altogether. For now, its not really worth moving to separate files or figuring out anything past what it there now... since it may ALL go away anyway.

docs/walkthrough/addNewDashboardPanel.md
elvinm: Yeah that does. I think I was thinking more of what is a page, vs a panel vs a tab but that would probably be in the first readme not here so feel free to ignore.

src/components/app/app.js
elvinm: That is fair enough

docs/walkthrough/addNewDashboardPanel.md
```diff
@@ -0,0 +1,78 @@
...
+- Technology reference:
+    - [react][react]
+    - [i18next][i18next]
+
+
+
+[ag-grid]: https://www.ag-grid.com/react-getting-started/
+[i18next]: https://www.i18next.com/
+[react]: https://reactjs.org/
+[redux]: https://redux.js.org/
```
elvinm: I don't see this or redux-obs anywhere when I preview

PR #1064, opened 2018-08-15
Add Rule Diagnostics

molly-moen:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

Add diagnostics logging for rule create/update events. Added the following metrics:
Rule_NewClick
Rule_EditClick
Rule_DeviceGroupClick
Rule_CalculationClick
Rule_FieldClick
Rule_OperatorClick
Rule_AddConditionClick
Rule_SeverityLevelClick
Rule_StatusToggle
Rule_ApplyClick
Rule_CancelClick
Rule_TopXCloseClick

Also includes new "sessionid" sections of diagnostics call, which logs the time in ms since Jan 1, 1970 when the page was loaded (amplitude expects session id in this format). This fields will be added by diagnostics to enable logging of session id to amplitude--until those changes go in it will be ignored by the backend. 

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/1064)

### DIFF
--- a/src/services/models/logEventModels.js
+++ b/src/services/models/logEventModels.js
@@ -4,7 +4,7 @@ import { toDiagnosticsModel } from 'services/models';
 
 export const toRuleDiagnosticsModel = (eventName, rule) =>
 {
-  var metadata = {
+  const metadata = {
     DeviceGroup: rule.groupId,
     Calculation : rule.calculation,
     TimePeriod: rule.timePeriod,
@@ -18,7 +18,6 @@ export const toRuleDiagnosticsModel = (eventName, rule) =>
 }
 
 export const toSinglePropertyDiagnosticsModel = (eventName, propertyTitle, property) => {
-  var metadata = {};
-  metadata[propertyTitle] = property;
+  const metadata = { [propertyTitle]: property };
   return toDiagnosticsModel(eventName, metadata);
 }
--- a/src/services/models/logEventModels.js
+++ b/src/services/models/logEventModels.js
@@ -2,11 +2,73 @@
 
 import { toDiagnosticsModel } from 'services/models';
 
-export const toLogRuleModel = (rule = {}) => ({
-  'DeviceGroup': rule.groupId,
-  'Calculation': rule.calculation,
-  'TimePeriod': rule.timePeriod,
-  'Severity': rule.severity
-});
+export const toLogRuleModel = (eventName, rule) =>
+{
+  var metadata = {
+    DeviceGroup: rule.groupId,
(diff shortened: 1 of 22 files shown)


src/store/reducers/appReducer.js
```diff
@@ -141,7 +144,17 @@ export const epics = createEpicScenario({
       GitHubService.getReleaseInfo()
         .map(toActionCreator(redux.actions.getReleaseInformation, fromAction))
         .catch(handleError(fromAction))
+  },
+
+  updateCurrentWindow: {
```
isaac73: change this into a reducer

src/store/reducers/appReducer.js
```diff
@@ -43,10 +44,12 @@ export const epics = createEpicScenario({
     epic: ({ payload }, store) => {
       const diagnosticsOptIn = getDiagnosticsOptIn(store.getState());
       if (diagnosticsOptIn) {
+        payload.sessionId = getSessionId(store.getState());
+        payload.eventProperties.CurrentWindow = getCurrentWindow(store.getState());
         return DiagnosticsService.logEvent(payload)
         /* We don't want anymore action to be executed after this call
             and hence return empty observable */
-          .map(_ => Observable.empty())
+          .flatMap(_ => Observable.empty())
```
isaac73: @spryor is this correct?

src/services/models/logEventModels.js
```diff
@@ -2,11 +2,23 @@
...
+    ConditionCount: rule.conditions.length,
+    FirstFieldChosen: rule.conditions[0].field,
+    FirstOperatorChosen: rule.conditions[0].operator
+  };
+
+  return toDiagnosticsModel(eventName, metadata);
+}
+
+export const toSinglePropertyDiagnosticsModel = (eventName, propertyTitle, property) => {
+  var metadata = {};
```
isaac73: @spryor is there a shortcut for this two line assignment?

spryor:
```js
const metadata = { propertyTitle: property };
```

molly-moen:
I think in order to keep the logging as flexible as possible for the future it would be best to not have generic logging in the form controls/state links. By keeping it within the editor it's very easy to add or remove extra metadata and be specific about what the rule and metadata names are. I agree that it would be nice to have an automated function, but I don't know a good way to do it without adding extra complexity and losing flexibility. I think it's also good to have all logging done in a consistent way.

PR #1069, opened 2018-08-18
Refactor to make walkthrough code less intrusive

mechaffin:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

1. Make current “app” page more of a “shell”
1. Separate walkthrough code into its own folder
1. Rename sample files with easier names
1. Create infrastructure so that the walkthrough code is almost completely separate from the real app.  The only cross over is in the main index.js page where it decides whether to load the regular “app” or the walkthrough “app.”

**Checklist:**

- [ ] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/1069)


src/walkthrough/components/app.js
```diff
@@ -0,0 +1,80 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+import React, { Component } from 'react';
+
+import { svgs } from 'utilities';
+import Shell from "components/shell/shell";
+import { DashboardContainer, BasicPageContainer, PageWithFlyoutContainer, PageWithGridContainer } from './pages';
+
+class App extends Component {
+  openSettings = () => {alert("There are no settings in the walkthrough.")};
```
spryor: Can just be 
```javascript
openSettings = () => alert('There are no settings in the walkthrough.');
```
Note the single quotes

src/components/app.js
```diff
@@ -0,0 +1,122 @@
...
+        to: '/maintenance',
+        exact: false,
+        svg: svgs.tabs.maintenance,
+        labelId: 'tabs.maintenance',
+        component: MaintenanceContainer
+      }
+    ];
+
+    const crumbsConfig = [
+      {
```
spryor: curmbsConfig should probably be combined with the pagesConfig

mechaffin: I really struggled with how to structure the breadcrumbs... and whether to combine with the pagesConfig.  For the top level addresses, they are very related and I didn't like the redundancy. However, for the other pages like "/maintenance/rule/:id" it felt odd to put those into pagesConfig with a boolean for includeInNav (or something similar).

src/components/app.js
```diff
@@ -0,0 +1,122 @@
...
+
+    const shellProps = {
+      pagesConfig,
+      crumbsConfig,
+      openSettings: this.openSettings,
+      ...this.props
+    };
+
+    return (
+      <Shell {...shellProps}>
```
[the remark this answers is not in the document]
mechaffin: If we come up with concrete fixes, we can certainly update as needed. For now, we should get this big set of changes in. Any modifications to the shell at this point should have a smaller footprint.

src/store/configureStore.js
```diff
@@ -6,7 +6,7 @@ import { composeWithDevTools } from 'redux-devtools-extension';
 import rootEpic from './rootEpic';
 import rootReducer from './rootReducer';
 
-export default function configureStore() {
```
isaac73: Why is default removed?

mechaffin: This was simply to have consistency between the normal configureStore and the one used in the walkthrough part of the code. They are both referenced in the `index.js` file and so at least one of them needs to be able to be renamed so we can refer to both in the same file...
```js
import { configureStore } from 'store/configureStore';
import { configureStore as configureWalkthroughStore } from 'walkthrough/store/configureStore';
```

PR #1075, opened 2018-08-28
Add diagnostics for new device funnel

molly-moen:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

Add metrics for creating a new device. Also add metric for when user turns on/off diagnostics. Also includes minor update for rule metric (emit apply event for edit and new rule, instead of just new rule)

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/1075)


src/components/pages/rules/flyouts/ruleEditor/ruleEditor.js
```diff
@@ -147,6 +147,7 @@ export class RuleEditor extends LinkedComponent {
           erroe: undefined
         }
       };
+      logEvent(toRuleDiagnosticsModel('Rule_ApplyClick', requestProps));
```
ppathan: Do we differentiate between create and update? Does ApplyClick is for both?

src/components/pages/devices/flyouts/deviceNew/deviceNew.js
```diff
@@ -242,6 +249,12 @@ export class DeviceNew extends LinkedComponent {
     ].every(link => !link.error);
   }
 
+  deviceTypeChange = ({ target: { value }}) => {
+    this.props.logEvent(toSinglePropertyDiagnosticsModel('Devices_DeviceTypeSelect', 'DeviceType',
+       (value ==='true') ? 'Simulated': 'Physical'));
```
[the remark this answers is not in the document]
molly-moen: added constant for simulated and physical. Enum already has true/false as values

src/components/pages/rules/flyouts/ruleEditor/ruleEditor.js
molly-moen: We have an initial click event that is different for edit/update but then all the events after are the same. We could differentiate here too but since the other events don't differentiate I thought it made more sense to keep it the same event

ppathan: That make sense.

Issue #1037, opened 2018-06-21
Jobs/Reconfigure not working as expected


bernieanger:
# Type of issue 

- [x ] Bug
- [ ] New feature
- [ ] Enhancement

# Description 
When issuing a "Reconfigure" Job on a physical device with a single Property of type Number the value set by the UI to the "DesiredProperties" section of the Digital Twin is set as a Text field (quoted value), even though it had been correctly identified as a Number in the UI

As a side effect, from that point on the property displays as "SyncingTo" ... as there will be a type mismatch 
...

# Steps to reproduce 

1. [First step]
Create a new physical device with a Digital Twin that includes a numeric property
2. [Second step]
In the RMD UI, schedule a job for the newly created device that Reconfigures the properties for the new device, and set a new value for the numeric property
 
3. [and so on...]
Submit the job & wait until it completes
# Expected behavior 
The desired property would be updated as a numerical value

# Current behavior 
...

# Possible solution 
...

# Context and Environment
* Operating System: ...
dot-net version, solution accelerator installed from the web ui

* GitHub branch/tag: ...
:testing (verified by looking into the VM)

# Screenshot 
Sample:

  "properties": {
    "desired": {
      "Latitude": 38.663,
      "Longitude": -78.4509,
      "CargoHumidity": 0,
      "CargoTemperature": 0,
      "Type": "Container21",
      "RPMSpeed": "3500",
      ...

    "reported": {
      "CargoHumidity": 0,
      "CargoTemperature": 0,
      "RPMSpeed": 3500,
      ...

PR #1082, opened 2018-08-31
Treat text and number correctly in device jobs

mechaffin:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Fix for https://github.com/Azure/pcs-remote-monitoring-webui/issues/1037.
Treat text and number correctly for the device jobs for tags and properties (reconfigure).

Also, fixed breadcrumb keys.

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/1082)


src/services/models/iotHubManagerModels.js
```diff
@@ -64,7 +64,8 @@ export const toSubmitTagsJobRequestModel = (devices, { jobName, updatedTags, del
   const jobId = jobName ? jobName + '-' + uuid() : uuid();
   const deviceIds = devices.map(({ id }) => `'${id}'`).join(',');
   const Tags = {};
-  updatedTags.forEach(({ name, value }) => (Tags[name] = value));
+  //Ensure type passed to server is correct as specified: number or text. The toString call is necessary when a number should be saved as text.
```
elvinm: Small nit: these lines are getting pretty long, can we format differently?

src/components/pages/devices/flyouts/deviceJobs/deviceJobTags.js
```diff
@@ -34,7 +34,7 @@ import {
 
 update.extend('$autoArray', (val, obj) => update(obj || [], val));
 
-const isNumeric = value => typeof value === 'number' || !isNaN(parseInt(value, 10));
+const isNumeric = value => typeof value === 'number';
```
elvinm: Is it guaranteed that isNumeric will only be used on values that are actually numbers? If a string value comes in that represents a number var num = "30", do we want that to also be a number? Which I think is why the !isNaN was there. Just want to make sure it isn't being used in this way.

mechaffin: Just because a value looks like a number doesn't mean that the actual type is a number. The tags and properties can be either text or number... and there is a drop down for the user to specify which one they want.

ppathan: BTW when use selects a number but enters a string do we check?

PR #1085, opened 2018-09-05
Add links to Time Series Insights from Dashboard and Device Details Page

jillcary:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

With the integration of Time Series Insights we are adding a link to view data for the solution on the Time Series Insights Explorer for the deployment.

This change includes the following:
- A Hyperlink shared component
- Fetch the Time Series Explorer URL from Telemetry
- Add link to dashboard and device details

**Checklist:**

- [ ] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/1085)


public/locales/en/translations.json
```diff
@@ -138,7 +138,8 @@
         "deviceTypeAlerts": "Alert by device type",
         "criticalAlerts": "Critical alerts",
         "currentWindow": "Currently",
-        "previousWindow": "Previously"
+        "previousWindow": "Previously",
+        "tsiExplorerUrl": "Explore in Time Series Insights"
```
mechaffin: These aren't URLs.  They're the friendly label for a URL. So something like tsiExplorerLinkLabel, but the suffix part isn't really needed (and we usually haven't put it in other translations)... so just plain tsiExplorer would be better... or maybe exploreTimeSeries.

Not sure I like using the TSI abbreviation, but won't fight too hard on it.

src/components/pages/devices/flyouts/deviceDetails/deviceDetails.js
```diff
@@ -174,6 +175,15 @@ export class DeviceDetails extends Component {
     const tags = Object.entries(device.tags || {});
     const properties = Object.entries(device.properties || {});
 
+    // Add parameters to Time Series Insights Url
+    const timeSeriesParamUrl =
+    timeSeriesExplorerUrl
+      ? timeSeriesExplorerUrl +
```
mechaffin: Consider more readable syntax (I'm sure there's a name for it, but I don't know what it is....) instead of concatenation.
Example: instead of 
```
'/maintenance/job/' + this.state.jobId
```
do this...
```
`/maintenance/job/${this.state.jobId}`
```

src/components/pages/dashboard/dashboard.js
```diff
@@ -317,6 +318,12 @@ export class Dashboard extends Component {
         ? deviceIds.length - onlineDeviceCount
         : undefined;
 
+    // Add parameters to Time Series Insights Url
+    const timeSeriesParamUrl =
+      timeSeriesExplorerUrl
+        ? timeSeriesExplorerUrl + '&relativeMillis=1800000&timeSeriesDefinitions=[{"name":"Devices","splitBy":"iothub-connection-device-id"}]'
```
mechaffin: See other comment about string concatenation.

Also, if these URL are complicated to build and similar to each other, would a helper function be useful?

src/components/shared/forms/hyperlink.js
```diff
@@ -0,0 +1,26 @@
...
+import { joinClasses } from 'utilities';
+
+import './styles/hyperlink.css';
+
+export const Hyperlink = (props) => {
+  const { children, className, href, target } = props;
+
+  if (href == null) return null;
+  return (
+    <a className={joinClasses('hyperlink', className)} href={href} target={target}>
```
mechaffin: I don't think you need to pull out href and target specifically.  They should get  passed on with {...props} right?
```html
<a className={joinClasses('hyperlink', className)} {...props}>
```

src/store/reducers/appReducer.js
```diff
@@ -76,6 +77,15 @@ export const epics = createEpicScenario({
         .catch(handleError(fromAction))
   },
 
+  /** Get Telemetry Properties */
+  fetchTelemetryProperties: {
+    type: 'APP_FETCH_TELEMETRY_PROPERTIES',
+    epic: (fromAction) =>
+      TelemetryService.getStatus()
```
mechaffin: Why the name mismatch?  properties vs status?

public/locales/en/translations.json
jillcary: Agree with the TSI abbreviation and the url label makes sense, I'll update it.

src/components/pages/dashboard/dashboard.js
jillcary: Opting to leave in as is. If we end up adding more complicated TSI queries we can revisit in a shared function in models perhaps.

Issue #1077, opened 2018-08-29
Rename flyouts which are inconsistent


isaac73:
# Type of issue 

- [x] Bug
- [ ] New feature
- [ ] Enhancement

# Description 

newRulesFlyout -> rulesNewFlyout
deleteRule.js -> ruleDelete.js
deleteRule.container.js -> ruleDelete.container.js
deleteRule.scss -> ruleDelete.scss
editRuleFlyout.js -> ruleEditFlyout.js

PR #1087, opened 2018-09-06
Rename Package Flyouts to be consistent with rest of UI

isaac73:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

...

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/1087)


isaac73:
https://github.com/Azure/pcs-remote-monitoring-webui/issues/1077

isaac73:
Initially, I thought of the same thing and that is why I created this issue #1077 . If you see the current naming conventions for flyouts: deviceDetails , deviceNew, deviceDelete, ruleNew, ruleDetails, ruleDelete, etc. this change aligns with it.

ppathan:
I would rather do #1077.

isaac73:
If we are looking for names that make sense, then any convention is not good. Consider this -> 
NewDevice makes sense but DetailsDevice doesn't.
DeviceDetails make sense but DeviceNew doesn't make sense.

So, I preferred sticking with the current convention.

src/components/pages/packages/flyouts/deletePackage/index.js
```diff
@@ -1,4 +0,0 @@
-// Copyright (c) Microsoft. All rights reserved.
```
ppathan: I think for all the renamed one's you can use git mv so it shows up as rename this will preserve history.

isaac73: Since these are just new flyouts in feature branch we don't need the history till now.

PR #1093, opened 2018-09-07
Rearrange controls on context menus

mechaffin:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

(TFS task 3006191)
Rearrange the controls per UX enhancements.
Left aligned: device group drown down & mange button.
Right aligned: everything else.
Also, move the time interval dropdown to the left of the refresh bar.

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/1093)


src/components/pages/dashboard/dashboard.js
```diff
@@ -342,17 +349,23 @@ export class Dashboard extends Component {
...
-          refresh={this.refreshDashboard}
-          time={lastRefreshed}
-          isPending={analyticsIsPending || devicesIsPending}
-          t={t} />
-        <TimeIntervalDropdown
-          onChange={this.props.updateTimeInterval}
-          value={timeInterval}
-          t={t} />
-        <ManageDeviceGroupsBtn />
+        <ContextMenuAlign key="left" left={true}>
```
elvinm: Can't this be simplified in the shared component? Everytime left {true} is passed we also pass key="left". Do we need both? Maybe I'm just slightly not understanding something here.

mechaffin: The key is required to help react differentiate between multiple of the same component "ContextMenuAlign" in its virtual DOM.... so it has to be on this component and not just added to the one underlying it. It doesn't have to be "left/right"... just a unique value on each (within its parent). I used left/right to help "document" which set it which... but if that makes it confusing any values would work.

PR #1095, opened 2018-09-12
Add glimmer icon to newly added grid items

mechaffin:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

1. Add default sort to the grids for devices, rules, and packages. 
1. Modify reducers for devices, rules, and packages to add "showGlimmer: true" when inserting a row in memory.
1. Add GlimmerRenderer.
1. Modify SoftSelectLinkCellRenderer to add the glimmer icon when showGlimmer === true.

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/1095)


src/components/shared/cellRenderers/cellRenderer.scss
```diff
@@ -35,6 +35,22 @@ $iconSize: 14px;
...
+    color: inherit;
+    text-decoration: underline;
+
+    &:hover { text-decoration: none; }
+  }
+
+  .glimmer-icon {
+    svg {
+      @include square-px-rem(10px);
+      @include rem-fallback(margin-left, -10px);
```
isaac73: May be, if you convert margin-left to padding left, the glimmer icon collapsing with other elements will be fixed.

Margins are considered to be outside of the element, and margins of adjacent items will overlap. The biggest difference between padding and margin is that vertical margins auto-collapse, and padding doesn't. This is why  padding is considered to be part of the element, and is always preserved

src/components/shared/cellRenderers/glimmerRenderer/glimmerRenderer.js
```diff
@@ -0,0 +1,14 @@
...
+import React from "react";
+
+import { svgs } from 'utilities';
+import { Svg } from 'components/shared/svg/svg';
+
+import '../cellRenderer.css';
+
+export const GlimmerRenderer = (props) => (
+  props.value
+  ? <Svg path={svgs.glimmer} className="glimmer-icon" />
```
isaac73: I think this ternary statement can be combined in single line.

src/components/shared/cellRenderers/softSelectLinkRenderer/softSelectLinkRenderer.js
```diff
@@ -16,11 +17,16 @@ export class SoftSelectLinkRenderer extends Component {
   };
 
   render() {
-    const { value, context } = this.props;
+    const { value, context, data } = this.props;
```
isaac73: where is 'data' coming from?

mechaffin: ag-grid passes it in

src/components/shared/cellRenderers/glimmerRenderer/glimmerRenderer.js
mechaffin: If I recall, it went over 90 characters.... and I didn't want you to ask me to split it into different lines. :)

src/components/shared/cellRenderers/cellRenderer.scss
mechaffin: I can play with it some more, but our goal was not to mess with the grid or its cells as they are today.  And the cells use margin (not padding now).  In order to accommodate the glimmer using padding, we would have to modify the cells to use it... which could impact its ability to auto size the columns.

src/components/shared/cellRenderers/cellRenderer.scss
```diff
@@ -56,27 +72,21 @@ $iconSize: 14px;
...
       .pcs-renderer-icon svg {
         fill: themed('colorCellRendererSvgFillHighlight');
 
         .disabled-icon { stroke: themed('colorCellRendererSvgFillHighlight'); }
       }
 
       .pcs-renderer-text { color: themed('colorCellRendererTextHighlight'); }
+    }
 
+    .glimmer-icon {
```
spryor: Since this only has one child you can inline this. `.glimmer-icon svg { fill: themed('colorGlimmerSvgFill'); }`

src/components/shared/cellRenderers/glimmerRenderer/glimmerRenderer.js
```diff
@@ -0,0 +1,14 @@
...
+
+import React from "react";
+
+import { svgs } from 'utilities';
+import { Svg } from 'components/shared/svg/svg';
+
+import '../cellRenderer.css';
+
+export const GlimmerRenderer = (props) => (
+  props.value
```
spryor: indentation nit:
```
 => (
  condition
    ? // true condition
    : // false condition
);
```

src/store/reducers/devicesReducer.js
```diff
@@ -81,7 +81,12 @@ const deleteDevicesReducer = (state, { payload }) => {
 };
 
 const insertDevicesReducer = (state, { payload }) => {
-  const { entities: { devices }, result } = normalize(payload, deviceListSchema);
+  const inserted = payload.reduce((acc, payloadItem) => {
```
spryor: Looks like payload is already an array so using a `map` here should be sufficient.
```js
const inserted = payload.map(device => ({ ...device, showGlimmer: true }));
```

spryor: The naming `showGlimmer` seems a bit too bound to the particular implementation of the real underlying idea - namely,  that the data is "new". I think a better name could be used in the data to indicate that the row is fresh or new.

src/store/reducers/rulesReducer.js
```diff
@@ -93,7 +93,12 @@ const ruleListSchema = new schema.Array(ruleSchema);
 const initialState = { ...errorPendingInitialState, entities: {}, items: [] };
 
 const insertRulesReducer = (state, { payload }) => {
-  const { entities: { rules }, result } = normalize(payload, ruleListSchema);
+  const inserted = payload.reduce((acc, payloadItem) => {
```
spryor: Again, looks like a map can be used here.

src/components/shared/cellRenderers/cellRenderer.scss
```diff
@@ -35,6 +35,22 @@ $iconSize: 14px;
...
 
+  .soft-select-link-cell {
+    display: inline;
+    color: inherit;
+    text-decoration: underline;
+
+    &:hover { text-decoration: none; }
+  }
+
+  .glimmer-icon {
```
spryor: `.glimmer-icon svg {`

PR #1104, opened 2018-09-20
More context bar arrangement; also page titles

mechaffin:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

1. All pages should have a Page Title, except the dashboard.
1. Refresh Bar should be placed on the page, except on the dashboard where it will remain on the context bar.
1. Refresh Bar should be right aligned above the page title.
1. Context Bar - Order of items aligned to the LEFT: device group dropdown, manage device group button
1. Context Bar - Order of items aligned to the RIGHT: search box, context buttons, time interval dropdown
1. Maintenance Alert details page - move the small stat grid under the page title (instead of allowing it to flow to side of the title)

**Checklist:**

- [x] All tests passed


src/components/pages/maintenance/jobDetails/jobDetails.js
```diff
@@ -46,10 +47,10 @@ export class JobDetails extends Component {
     //       A long term fix would be to normalize the job data in maintenance.js (similar to how Telemetry is handled there).
     //       When/if that happens, remove all use of refreshPending in the local state of this component.
     if ((
-        nextProps.match.params.id !== (this.state.selectedJob || {}).jobId
-        || this.state.timeIntervalChangePending
-        || this.state.refreshPending
-      ) && nextProps.jobs.length) {
+      nextProps.match.params.id !== (this.state.selectedJob || {}).jobId
```
isaac73: Alignment could be wrong. Also, if the above open brackets are in separate lines it will be clearer so visualize the block

src/components/pages/maintenance/jobDetails/jobDetails.js
```diff
@@ -126,31 +127,31 @@ export class JobDetails extends Component {
...
-                {
-                  this.state.selectedDevices
-                    ? <DevicesGrid
-                        t={t}
-                        domLayout={'autoHeight'}
-                        rowData={this.state.selectedDevices}
-                        onContextMenuChange={this.onContextMenuChange} />
-                    : t('maintenance.noOccurrenceSelected')
-                }
+              <JobGrid {...jobGridProps} />
```
isaac73: alignment

src/components/pages/maintenance/ruleDetails/ruleDetails.js
```diff
@@ -304,19 +305,19 @@ export class RuleDetails extends Component {
...
-            t={t} />
         </ContextMenuAlign>
       </ContextMenu>,
       <PageContent className="maintenance-container rule-details-container" key="page-content">
+      <RefreshBar
+          refresh={this.refreshData}
+          time={lastUpdated}
+          isPending={isPending}
+          t={t} />
+        <PageTitle titleValue={alertName} />
```
isaac73: check alignment

mechaffin:
Updated formatting for those 2 files using Alt-Shft-F. In order to spacing to look correct with this auto formatting (which is really the only way to ensure consistency), some opening element tags were moved to the next line. Without that, the opening tag and the attributes on the subsequent lines end up in the same column... which is not how we want it to look. You'll notice there was some other formatting/indentation that was changed, but again... using Alt-Shft-F is really the only way to ensure consistency. We have project workspace setttings in this project to ensure we're all using the same settings.

Issue #1079, opened 2018-08-30
UI errors out if tab is left open


isaac73:
# Type of issue 

- [x] Bug
- [ ] New feature
- [ ] Enhancement

# Description 

UI errors out if tab is left open

![image](https://user-images.githubusercontent.com/22013837/44870212-6f4ef800-ac44-11e8-99a4-70b545a3a657.png)

# Steps to reproduce 

1. open remote monitoring UI (Dashborad page)
2. Do not close the tab for long time (I think around 30 to 1 hour)
3. screen errors out

# Expected behavior 
...

# Current behavior 
...

# Possible solution 
...

# Context and Environment
* Operating System: ...
* GitHub branch/tag: ...

# Screenshot

asdonald:
@asdonald to start a thread on this.

asdonald:
@mechaffin to look into for Oct.

asdonald:
@mechaffin Moving to 1811

mechaffin:
Any additional fixes should be P3 until and unless further issues are seen after the initial fix.

PR #1147, opened 2018-10-26
Get Auth Token in an Async way to resolve dashboard Oops error when idle

mechaffin:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Change handling of getting the auth token to be async. This should fix the "oops" errors that occur on the dashboard when left idle for a long period of time (and the token times out and gets refreshed).

Addresses: https://github.com/Azure/pcs-remote-monitoring-webui/issues/1079

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/1147)


src/services/authService.js
```diff
@@ -128,22 +128,23 @@ export class AuthService {
    * Acquires token from the cache if it is not expired.
    * Otherwise sends request to AAD to obtain a new token.
    */
-  static getAccessToken(callback) {
+  static getAccessToken() {
     if (AuthService.isDisabled()) {
-      if (callback) callback('client-auth-disabled');
-      return;
+      return Observable.of('client-auth-disabled');
```
ppathan: Did moving to Observable.of() fixed the issue for the HttpClient change?

PR #1150, opened 2018-11-02
Fix rule enable/disable bug

molly-moen:
# Type of change? 
- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Could not enable/disable rules without actions from status flyout due to null error. Also fixed issue of disabled rules showing as enabled in summary flyout.

**Checklist:**

- [x] All tests passed


src/services/models/telemetryModels.js
```diff
@@ -147,14 +147,16 @@ export const toEditRuleRequestModel = ({
...
 
-  const Actions = actions.map(act => ({
-    Type: act.type,
-    Parameters: {
-      Notes: act.parameters.notes,
-      Recipients: act.parameters.recipients,
-      Subject: act.parameters.subject
-    }
-  }));
+  const Actions = actions
```
mechaffin: const Actions = (actions || []).map(act => ({
    Type: act.type,
    Parameters: {
      Notes: act.parameters.notes,
      Recipients: act.parameters.recipients,
      Subject: act.parameters.subject
    }
  }));

PR #1163, opened 2018-11-08
Update Outlook login flow

molly-moen:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

Only show outlook login flow for applications that were deployed as owners, because those that are not deployed as owners cannot check that login was completed. Therefore we always have to show the same information whether or not login completed for non-owners. In addition, only show "setup complete" immediately after setup is clicked. Also add diagnostics logging specific to actions.
**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/1163)


src/components/pages/rules/rules.js
```diff
@@ -41,13 +42,20 @@ export class Rules extends Component {
     }
 
     this.props.updateCurrentWindow('Rules');
+    if (this.props.isDeployedAsOwner !== undefined) {
```
mechaffin: if (!this.props.isDeployedAsOwner) {

src/components/pages/rules/rules.js
```diff
@@ -69,6 +77,15 @@ export class Rules extends Component {
 
   onContextMenuChange = contextBtns => this.setState({ contextBtns });
 
+  logApplicationPermissions(isOwner) {
+    if (isOwner !== undefined) {
```
mechaffin: if (!isOwner) {

src/store/reducers/appReducer.js
```diff
@@ -430,4 +434,6 @@ export const getActionPollingError = state =>
   getError(getAppReducer(state), epics.actionTypes.pollActionSettings);
 
 export const getActionPollingTimeout = state => getAppReducer(state).actionPollingTimeout;
+
+export const isDeployedAsOwner = state => getAppReducer(state).isDeployedAsOwner;
```
mechaffin: Aren't all the other selectors named with "get" at the beginning?

src/components/pages/rules/flyouts/ruleEditor/actionEmailSetup.js
```diff
@@ -40,10 +52,14 @@ export class ActionEmailSetup extends Component {
...
+    // Only show setup information if the application was deployed as owner,
+    // as if it was not cannot verify whether or not setup is complete.
+    // Also only show verification of setup if setup button was clicked.
     return (
-      <div className="action-email-setup-container">
+      isDeployedAsOwner
+      ? <div className="action-email-setup-container">
         {
-          showPending &&
+          showPending && clickedSetup &&
```
mechaffin: Couldn't you add clickedSetup to the calculation for showPending?

src/components/pages/rules/flyouts/ruleEditor/actionEmailSetup.js
```diff
@@ -86,13 +102,14 @@ export class ActionEmailSetup extends Component {
           </div>
         }
         {
-          showSetupComplete &&
+          showSetupComplete && clickedSetup &&
```
mechaffin: Couldn't you add clickedSetup to the calculation for showSetupComplete?

src/components/pages/rules/rules.js
molly-moen: I do want to log if it's false

src/store/reducers/appReducer.js
molly-moen: for boolean we have isDefaultLogo. I'm open to changing it, but since it return true/false isDeployedAsOwner makes more sense to me. Otherwise it would be getIsDeployedAsOwner

src/components/pages/rules/flyouts/ruleEditor/actionEmailSetup.js
molly-moen: good call

src/store/reducers/appReducer.js
```diff
@@ -282,8 +284,10 @@ const updateSolutionSettingsReducer = (state, { payload, fromAction }) => update
 
 const updateActionSettingsReducer = (state, { payload, fromAction }) => {
   const { entities: { actionSettings } } = normalize(payload, actionSettingsListSchema);
+  const isDeployedAsOwner = actionSettings.Email.applicationPermissionsAssigned;
```
zhang-hua: nit: The name maybe better to be consistent with back-end to say 'applicationPermissionsAssigned' which is not assuming 'owner' to deploy this solution. it could be other similar admin role have permission to assign role for application.

molly-moen: done

src/store/reducers/appReducer.js
molly-moen: I've renamed the variable so this is now a get

src/store/reducers/appReducer.js
```diff
@@ -282,8 +284,13 @@ const updateSolutionSettingsReducer = (state, { payload, fromAction }) => update
 
 const updateActionSettingsReducer = (state, { payload, fromAction }) => {
   const { entities: { actionSettings } } = normalize(payload, actionSettingsListSchema);
+  var applicationPermissionsAssigned = undefined;
```
mechaffin: Consider using 
```js
dot.pick('Email.applicationPermissionsAssigned', actionSettings);
```
to avoid the null check. Reduces lines of code. Similar usage in `pollActionSettings` in this same file.

PR #1164, opened 2018-11-09
Simple Tooltip component plus TSI explorer & rules/actions tooltips

mechaffin:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

1. Created a simple Tooltip component (replaces ContextHelpLink).
1. Use new Tooltip in TSI explorer links as well as Rules/Actions setup.
1. Add new checkmark bubble icon (and remove old one).

**Checklist:**

- [x] All tests passed


src/components/pages/rules/flyouts/ruleEditor/ruleEditor.js
```diff
@@ -40,8 +46,7 @@ import {
   toDiagnosticsModel,
   toSinglePropertyDiagnosticsModel
 } from 'services/models';
-import Config from 'app.config';
-import update from 'immutability-helper';
+
```
molly-moen: nit: extra blank line

src/components/shared/forms/styles/tooltip.scss
```diff
@@ -0,0 +1,111 @@
...
+      bottom: auto;
+      left: 50%;
+      right: auto;
+    }
+  }
+
+  @include themify($themes) {
+    .tooltip {
+      .tooltip-content {
+        color:  themed('colorTooltipText');
```
spryor: multiple spaces after :

src/components/shared/forms/styles/tooltip.scss
```diff
@@ -0,0 +1,111 @@
...
+        background-color:  themed('colorTooltipBackground');
+
+        &:before,
+        &:after {
+          background-color:  themed('colorTooltipBackground');
```
spryor: here too

src/components/shared/forms/tooltip.js
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+import React, { Component } from 'react';
+
+import { joinClasses } from 'utilities';
+
+import './styles/tooltip.css';
+
+export class Tooltip extends Component {
```
spryor: Make this a functional component

src/components/shared/forms/tooltip.js
```diff
@@ -0,0 +1,25 @@
...
+  render() {
+    const { target, children, position } = this.props;
+
+    return (
+      <div class="tooltip-container">
+        <div class={joinClasses('tooltip', position)}>
+          <div class="tooltip-content">
+            {children}
+          </div>
+          <div class="tooltip-arrow"></div>
```
spryor: nit: since this is jsx and not html your can do this `<div class="tooltip-arrow" />`

src/components/shared/forms/tooltip.js
```diff
@@ -0,0 +1,25 @@
...
+import { joinClasses } from 'utilities';
+
+import './styles/tooltip.css';
+
+export class Tooltip extends Component {
+  render() {
+    const { target, children, position } = this.props;
+
+    return (
+      <div class="tooltip-container">
```
spryor: These should be className

src/components/shared/forms/styles/tooltip.scss
```diff
@@ -0,0 +1,111 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+@import 'src/styles/themes';
+@import 'src/styles/mixins';
+
+.tooltip-container {
+  @include rem-font-size(12px);
```
spryor: The pattern is that `@includes`  should go at the end of the list of properties

src/components/shared/forms/tooltip.js
```diff
@@ -0,0 +1,25 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+import React, { Component } from 'react';
```
spryor: It feels a bit odd to pass the main content (the stuff that causes a tooltip to appear) as a prop instead of as the children. This would feel more natural.

```jsx
<Tooltip content={getTooltip()}>
  <div>
    Some stuff that will show a tooltip if hovered over.
  </div>
</Tooltip>
```

Other react UI libs like Material and Bootstrap use this pattern too.

src/components/pages/rules/flyouts/ruleEditor/ruleEditor.js
mechaffin: Actually, its a convention used throughout the application that we put a blank line between the 3rd party stuff and our stuff in the import lists.

src/components/shared/forms/tooltip.js
mechaffin: Well heck, React isn't doing us any favors, is it?  It used to not work if you messed that up.

src/components/shared/forms/styles/tooltip.scss
mechaffin: Hmmm, not sure how I got it in my head they were supposed to be first.

src/components/shared/forms/tooltip.js
mechaffin: Most of the react components I researched did it the way I have it actually. To me it made more sense for the contents of the Tooltip (which is the name of the component) to be the children of that component.  Also, the stuff inside a tooltip (the way our designers have it) is often more complicated than the thing that triggers it.

mechaffin: Material UI does as you say. However, bootstrap does not. In that case the tooltip is just the toolip and they have an OverlayTrigger component where the tooltip is a prop.

mechaffin: For the sake of getting it done and since its all likely to change in Vanadium, I'd say let's leave it as is. If you feel strongly otherwise, let's talk in person.

spryor: The material tooltip is named differently but has the same behavior as the bootstrap version with overlay. So the factoring is a bit different but the implementation (and end usage) is essentially the same. The material design is also more opinionated about what a tooltip is so that allows it to get away with less customization than bootstrap. I think the tooltip content as a prop vs children is the more logical approach overall and given that that is also the design choice of those major libraries I think their years of experience in the wild testifies to that.

PR #1168, opened 2018-11-14
User Profile flyout to show roles & permissions

mechaffin:
# Type of change? 

- [ ] Bug fix
- [x] New feature
- [ ] Enhancement

Remove old menu under the user icon on the header. Replace with a flyout displaying their roles and permissions. Also, the logout button to be located on the flyout.

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/1168)


src/app.config.js
```diff
@@ -16,6 +16,7 @@ const Config = {
   },
   contextHelpUrls: {
     accessDenied: 'https://docs.microsoft.com/en-us/azure/iot-accelerators/iot-accelerators-remote-monitoring-rbac#add-or-remove-users',
+    rolesAndPermissions: 'https://docs.microsoft.com/en-us/azure/iot-accelerators/iot-accelerators-remote-monitoring-rbac',
     ruleActionsEmail: 'https://go.microsoft.com/fwlink/?linkid=2041110&clcid=0x409',
     exploreTimeSeries: 'https://docs.microsoft.com/en-us/azure/iot-accelerators/iot-accelerators-remote-monitoring-rbac-tsi'
```
ppathan: I see all the links have en-us in them. Please remove them.

src/components/shell/flyouts/profile/profile.js
```diff
@@ -0,0 +1,111 @@
...
+        <Flyout.Title>{t('profileFlyout.title')}</Flyout.Title>
+        <Flyout.CloseBtn onClick={onClose} />
+      </Flyout.Header>
+      <Flyout.Content className="profile-container">
+        {
+          !user &&
+          <div className="profile-container">
+            <ErrorMsg className="profile-error">{t("profileFlyout.noUser")}</ErrorMsg>
+            <Trans i18nKey={'profileFlyout.description'}>
+              <Hyperlink href={Config.contextHelpUrls.rolesAndPermissions} target="_blank">{t('profileFlyout.learnMore')}</Hyperlink> about roles and permisions
```
ppathan: At the end of this line there is some text, seems like a copy/paste error?

ppathan: Ahh looks like  the text is already covered in the translations so it can be removed?

src/components/shell/flyouts/profile/profile.js
```diff
@@ -0,0 +1,111 @@
...
+          user &&
+          <div className="profile-container">
+            <div className="profile-header">
+              <h2>{user.email}</h2>
+              <Grid className="profile-header-grid">
+                <Row>
+                  <Cell className="col-7">
+                    <Trans i18nKey={'profileFlyout.description'}>
+                      <Hyperlink href={Config.contextHelpUrls.rolesAndPermissions} target="_blank">{t('profileFlyout.learnMore')}</Hyperlink>
+                      about roles and permisions
```
ppathan: remove here as well?

src/components/shell/flyouts/profile/profile.js
```diff
@@ -0,0 +1,111 @@
...
+                          </Row>
+                        )
+                      }
+                    </Grid>
+                }
+              </Section.Content>
+            </Section.Container>
+
+            <Section.Container>
+              <Section.Header>{t('profileFlyout.permissions')}</Section.Header>
```
ppathan: Q: Will the user able to figure out from UI what permissions give them what kind of roles?

src/services/models/authModels.js
```diff
@@ -29,13 +29,14 @@ export const permissions = {
   createPackages: 'CreatePackages',
 
   createDeployments: 'CreateDeployments',
-  delteDeployments: 'DeleteDeployments'
+  deleteDeployments: 'DeleteDeployments'
```
ppathan: This reference wasn't used any where else? Looks like this fixes a bug in perms :)

src/components/shell/flyouts/profile/profile.js
mechaffin: There has to be some sort of text there as it sets the template for what it gets replaced with.

src/components/shell/flyouts/profile/profile.js
mechaffin: No, its needed for the template

src/services/models/authModels.js
mechaffin: It wasn't used anywhere (yet), but it was misspelled... so I fixed it.

src/components/shell/flyouts/profile/profile.js
mechaffin: The role is what gives them individual permissions.  But no, they cannot tell which role grants each permission.  Currently, the backend gives us an aggregate of the permissions from all roles.... so a simple string array list of the permissions they have and no association to which role (as it could be several) that granted it.

PR #1178, opened 2018-11-20
Use POST Request for long queries

molly-moen:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

Telemetry get request urls can be longer than 2048 characters if have many devices. Send a POST request instead for these types of queries. Also remove devices from iot hub manager jobs query, as jobs cannot be filtered on devices and this query failed if the url got too long.
Related to this issue: https://github.com/Azure/azure-iot-pcs-remote-monitoring-dotnet/issues/108

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/1178)

### DIFF
--- a/src/services/telemetryService.js
+++ b/src/services/telemetryService.js
@@ -45,20 +45,17 @@ export class TelemetryService {
 
   /** Returns a list of alarms (all statuses) */
   static getAlerts(params = {}) {
-    return HttpClient.get(`${ENDPOINT}alarms?${stringify(params)}`)
-      .map(toAlertsModel)
+    return this.getOrPost(`${ENDPOINT}alarms`, params, toAlertsModel);
   }
 
   /** Returns a list of active alarms (open or ack) */
   static getActiveAlerts(params = {}) {
-    return HttpClient.get(`${ENDPOINT}alarmsbyrule?${stringify(params)}`)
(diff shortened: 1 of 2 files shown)


src/services/telemetryService.js
```diff
@@ -105,4 +102,20 @@ export class TelemetryService {
       .map(() => ({ deletedRuleId: id }));
   }
 
+  static getOrPost(url, params, toModel, devicesArr = undefined) {
+    const urlWithParams = `${url}?${stringify(params)}`;
+    if (urlWithParams.length > 2048) {
```
ppathan: nit: Make 2048 a const. Also add a comment as to why we are doing it this way.

src/services/telemetryService.js
```diff
@@ -105,4 +102,20 @@ export class TelemetryService {
...
+      if (devicesArr) {
+        params.devices = devicesArr;
+      } else if (params.devices && !Array.isArray(params.devices)) {
+        params.devices = params.devices.split(",");
+      }
+      return HttpClient.post(url, params)
```
ppathan: Should we just call post all the time?

PR #1194, opened 2018-12-05
Add Diagnostics Events

molly-moen:
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [x] Enhancement

Add diagnostic events for the following categories:
-Overall dashboard events
-Device group create/edit flow
-Rules page open
-Help dropdown
-Refresh 

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/1194)


src/components/pages/dashboard/dashboard.js
```diff
@@ -400,7 +402,8 @@ export class Dashboard extends Component {
                   mapKeyIsPending={azureMapsKeyIsPending}
                   isPending={devicesIsPending || analyticsIsPending}
                   error={azureMapsKeyError || devicesError || analyticsError}
-                  t={t} />
+                  t={t}
+                  logEvent = {logEvent} />
```
mechaffin: extra spaces?

src/components/pages/dashboard/dashboard.js
```diff
@@ -409,7 +412,8 @@ export class Dashboard extends Component {
                 isPending={analyticsIsPending || rulesIsPending}
                 error={rulesError || analyticsError}
                 t={t}
-                deviceGroups={deviceGroups} />
+                deviceGroups={deviceGroups}
+                logEvent = {logEvent} />
```
mechaffin: extra spaces?

src/components/pages/dashboard/panels/alerts/alertsPanel.js
```diff
@@ -14,6 +14,10 @@ import {
 } from 'components/pages/dashboard/panel';
 import { RulesGrid, rulesColumnDefs } from 'components/pages/rules/rulesGrid';
 import { translateColumnDefs } from 'utilities';
+import { toDiagnosticsModel } from 'services/models';
+import {
```
mechaffin: This import should be on a single line.

src/components/pages/dashboard/panels/alerts/alertsPanel.js
```diff
@@ -31,10 +35,17 @@ export class AlertsPanel extends Component {
         headerName: 'rules.grid.count',
         field: 'count'
       },
-      rulesColumnDefs.explore
+      {
+        ...rulesColumnDefs.explore,
+        cellRendererFramework: props => <LinkRenderer {...props} to={`/maintenance/rule/${props.value}`} onLinkClick={this.logExploreClick} />
```
mechaffin: Why not add this in the xxConfig.js file? Don't you want to log whenever this link is clicked?

src/components/shared/refreshBar/refreshBar.js
```diff
@@ -3,14 +3,18 @@
...
 import { DEFAULT_TIME_FORMAT } from 'components/shared/pcsGrid/pcsGridConfig';
 
 import './refreshBar.scss';
 
 export class RefreshBar extends Component {
 
-  refresh = () => !this.props.isPending && this.props.refresh();
+  refresh = () => {
+    this.props.logEvent(toDiagnosticsModel('Refresh_Clicked', {}));
+    !this.props.isPending && this.props.refresh();
```
mechaffin: return? The old way would return the boolean result.  This way will not.

src/components/shell/flyouts/manageDeviceGroups/manageDeviceGroups.js
```diff
@@ -45,25 +45,38 @@ export class ManageDeviceGroups extends LinkedComponent {
...
     addNewDeviceGroup: false,
     selectedDeviceGroup: undefined
   });
 
-  onEditDeviceGroup = selectedDeviceGroup => () => this.setState({
-    selectedDeviceGroup
-  });
+  onEditDeviceGroup = selectedDeviceGroup => () => {
+    this.props.logEvent(toDiagnosticsModel('DeviceGroup_EditClick', {}));
+    this.setState({
```
mechaffin: one line?

src/components/shell/header/header.container.js
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+import { connect } from 'react-redux';
+import { translate } from 'react-i18next';
```
mechaffin: Do you need translate?  If Header was relying on the t() function before, where was it getting it from? Do those locations need to be cleaned up?

src/components/shared/refreshBar/refreshBar.container.js
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+import { connect } from 'react-redux';
+import { translate } from 'react-i18next';
```
mechaffin: Do you need translate?  If Header was relying on the t() function before, where was it getting it from? Do those locations need to be cleaned up?

src/components/pages/dashboard/panels/telemetry/telemetryChart.container.js
```diff
@@ -0,0 +1,13 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+import { connect } from 'react-redux';
+import { translate } from 'react-i18next';
```
mechaffin: Do you need translate?  If Header was relying on the t() function before, where was it getting it from? Do those locations need to be cleaned up?

mechaffin: Also note, with my earlier checkin... we won't use translate() anymore, instead use withNamespaces(). Its the same thing, they just renamed it in a later version.

src/components/shared/refreshBar/refreshBar.container.js
mechaffin: Also note, with my earlier checkin... we won't use translate() anymore, instead use withNamespaces(). Its the same thing, they just renamed it in a later version.

src/components/shell/flyouts/manageDeviceGroups/views/deviceGroupForm.js
```diff
@@ -122,15 +123,22 @@ class DeviceGroupForm extends LinkedComponent {
...
     newCondition()
-  ]);
+    ]);
+  }
 
   deleteCondition = (index) =>
-    () => this.conditionsLink.set(this.conditionsLink.value.filter((_, idx) => index !== idx));
+    () => {
+      this.props.logEvent(toDiagnosticsModel('DeviceGroup_RemoveCondition', {}));
+      this.conditionsLink.set(this.conditionsLink.value.filter((_, idx) => index !== idx));
```
mechaffin: Does the code rely on this returning anything? The old code did return something; the new doesn't.

src/components/pages/dashboard/panels/alerts/alertsPanel.js
molly-moen: discussed offline

src/components/pages/dashboard/panels/telemetry/telemetryChart.container.js
molly-moen: you're right, didn't need

mechaffin: Looks like its still there.... and named translate instead of withNamespaces.

src/components/shared/refreshBar/refreshBar.container.js
mechaffin: Looks like translate is still there.... and named translate instead of withNamespaces. Is it needed here?

src/components/shell/header/header.container.js
```diff
@@ -0,0 +1,12 @@
...
+import { connect } from 'react-redux';
+
+import { epics } from 'store/reducers/appReducer';
+import { Header } from './header';
+
+const mapDispatchToProps = dispatch => ({
+  logEvent: diagnosticsModel => dispatch(epics.actions.logEvent(diagnosticsModel))
+});
+
+export const HeaderContainer = (connect(null, mapDispatchToProps)(Header));
```
mechaffin: nit: extra parens

src/components/shell/timeIntervalDropdown/timeIntervalDropdown.container.js
```diff
@@ -0,0 +1,12 @@
+// Copyright (c) Microsoft. All rights reserved.
+
+import { connect } from 'react-redux';
+import { translate } from 'react-i18next';
```
mechaffin: Is translate (please change to withNamespaces) needed here? If so, did it already rely on t being passed as a prop? Have those places been cleaned up now that the container supplies it?

molly-moen:
adding Context API code might be out of scope for this PR. Will investigate further

src/components/pages/dashboard/dashboard.js
```diff
@@ -400,7 +402,8 @@ export class Dashboard extends Component {
                   mapKeyIsPending={azureMapsKeyIsPending}
                   isPending={devicesIsPending || analyticsIsPending}
                   error={azureMapsKeyError || devicesError || analyticsError}
-                  t={t} />
+                  t={t}
+                  logEvent={logEvent} />
```
sakimsft: What is the reason passing logEvent from dashboard to each panel instead of each panel itself have a container and dispatch logEvent?
Although logEvent does not have payload in this case, we may want to add metadata for each panel later and these metadata belong to panel's state rather than dashboard's state. 
Also, if the component is reused by other pages, other pages should also implement and pass logEvent to each components. (this is not the case for a panel because it is not shared currently, but I think having a container for a panel would be preferable pattern)

src/components/pages/dashboard/dashboard.js
```diff
@@ -409,7 +412,8 @@ export class Dashboard extends Component {
                 isPending={analyticsIsPending || rulesIsPending}
                 error={rulesError || analyticsError}
                 t={t}
-                deviceGroups={deviceGroups} />
+                deviceGroups={deviceGroups}
+                logEvent={logEvent} />
```
sakimsft: Same comment as above.

src/components/pages/rules/rules.js
```diff
@@ -43,6 +43,8 @@ export class Rules extends Component {
     }
 
     this.props.updateCurrentWindow('Rules');
+    this.props.logEvent(toDiagnosticsModel('RulesPage_Click', {}));
```
sakimsft: Same comment as above.
You can remove this event if you want since I implemented this from navigation menu for all tabs.

src/components/shared/refreshBar/refreshBar.js
```diff
@@ -3,14 +3,18 @@
...
 import { svgs } from 'utilities';
 import { DEFAULT_TIME_FORMAT } from 'components/shared/pcsGrid/pcsGridConfig';
 
 import './refreshBar.scss';
 
 export class RefreshBar extends Component {
 
-  refresh = () => !this.props.isPending && this.props.refresh();
+  refresh = () => {
+    this.props.logEvent(toDiagnosticsModel('Refresh_Clicked', {}));
```
sakimsft: nit: most event names end "Click" instead of "Clicked".  rename "Click" for consistency.

src/components/shell/deviceGroupDropdown/deviceGroupDropdown.js
```diff
@@ -3,12 +3,14 @@
...
 
 import { Select } from 'components/shared';
+import { toDiagnosticsModel } from 'services/models';
 
 import './deviceGroupDropdown.scss';
 
 export class DeviceGroupDropdown extends Component {
 
   onChange = (deviceGroupIds) => ({ target: { value: { value } = {} } = {} }) => {
+    this.props.logEvent(toDiagnosticsModel('DeviceGroupFilter_Selected', {}));
```
sakimsft: nit: event names end "Select" instead of "Selected"  rename to "Select"

src/components/shell/timeIntervalDropdown/timeIntervalDropdown.js
```diff
@@ -17,6 +18,7 @@ const optionValues = [
 export class TimeIntervalDropdown extends Component {
 
   onChange = (propOnChange) => ({ target: { value: { value } = {} } = {} }) => {
+    this.props.logEvent(toDiagnosticsModel('TimeFilter_Selected', {}));
```
sakimsft: nit: rename Selected to Select.

src/components/pages/dashboard/dashboard.js
molly-moen: the reason was to simplify the code because we are only using the panels from the dashboard, so it was simpler to not add extra container files. I don't have a strong preference either way, but for components that were only used once this seemed simpler

molly-moen: since I'm taking out the call to log event from the dashboard code I will add containers

PR #1197, opened 2018-12-07
Fix up styles with ellipsis & add tooltips for deployment columns

mechaffin:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

1. Fix up styles with ellipsis: grid columns, stat values, config type dropdown
1. Add tooltips for deployment columns: Applied, Targeted

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/1197)


src/components/shared/cellRenderers/timeRenderer/timeRenderer.js
```diff
@@ -16,6 +18,12 @@ const formatTime = (value) => {
 export const TimeRenderer = ({ value }) => {
   const formattedTime = formatTime(value);
   return (
-    formattedTime ? formattedTime : EMPTY_FIELD_VAL
+    <div className="pcs-renderer-cell">
```
sakimsft: If we want to style error text in a same way, don't we need to wrap empty field value as well?
Before, it was all string values and now TimeRenderer returns div tag while empty value is still remains as string at below.

lastTriggerRenderer.js line 14, 
 ? value.error ? EMPTY_FIELD_VAL : <TimeRenderer value={value.response} /> 

ruleDetails.js line 348, Config.emptyValue

mechaffin: The point is to add ellipsis when needed. The error case will just be showing "--" so the ability to do ellipsis there isn't really necessary.

PR #1203, opened 2018-12-15
Bug fix for Alerts on device details page

sushilraje:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

The alerts section on the device details page flyouts was showing ALL alerts for all the device(s) rather than showing alerts for the particular device only. The issue was traced to the POST calls to Telemetry service. The call body contained JSON attributes in lower case, but JAVA JsonProperty mapping expected Pascal case attributes. To make the calls consistent with mappings (QueryApiModel) the attributes, 
1. Devices
2. From
3. To
4. Limit
5. Order

are converted into Pascal case. This does not break the dotnet version.

**Checklist:**

- [x] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/1202)

### DIFF
--- a/src/services/telemetryService.js
+++ b/src/services/telemetryService.js
@@ -11,7 +11,8 @@ import {
   toMessagesModel,
   toRuleModel,
   toRulesModel,
-  toStatusModel
+  toStatusModel,
+  toTelemetryRequestModel
 } from './models';
 
 const ENDPOINT = Config.serviceUrls.telemetry;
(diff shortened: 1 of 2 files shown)


src/services/telemetryService.js
```diff
@@ -48,7 +49,7 @@ export class TelemetryService {
     if (params.devices && !Array.isArray(params.devices)) {
       params.devices = params.devices.split(",");
     }
-    return HttpClient.post(`${ENDPOINT}alarms`, params)
+    return HttpClient.post(`${ENDPOINT}alarms`, toTelemetryRequestModel(params))
```
mechaffin: Call this before calling the service like everywhere else.

PR #1204, opened 2019-01-10
Read message schema from module twins

elvinm:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

When edge devices are reporting telemetry there is currently no way to create rules against that data. This is because the rules flyout only looks in the device twin for message schema. 

This PR makes a query for modules in the same device group to see if they are reporting a message schema in the module twin. The expected schema to use is something like:

**Checklist:**

- [ ] All tests passed

### DIFF
--- a/src/components/pages/rules/flyouts/ruleEditor/ruleEditor.js
+++ b/src/components/pages/rules/flyouts/ruleEditor/ruleEditor.js
@@ -259,11 +259,20 @@ export class RuleEditor extends LinkedComponent {
       if (group.id === groupId) {
         if (this.subscription) this.subscription.unsubscribe();
         this.subscription = IoTHubManagerService.getDevices(group.conditions)
+          .flatMap(devices => {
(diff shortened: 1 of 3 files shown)


src/components/pages/rules/flyouts/ruleEditor/ruleEditor.js
```diff
@@ -259,11 +259,20 @@ export class RuleEditor extends LinkedComponent {
       if (group.id === groupId) {
         if (this.subscription) this.subscription.unsubscribe();
         this.subscription = IoTHubManagerService.getDevices(group.conditions)
+          .flatMap(devices => {
+            const deviceIds = devices.map(dvc => `'${dvc.id}'`).join(",");
+            const modulesQuery = `deviceId IN [${deviceIds}]`;
+            return IoTHubManagerService.getModulesFields(modulesQuery);
+          }, (devices, modules) => {
+            return [devices, modules];
```
sushilraje: Should this be an associative array?

PR #1216, opened 2019-02-12
Using 'value' for selectField

isaac73:
# Type of change? 

- [x] Bug fix
- [ ] New feature
- [ ] Enhancement

DeviceGroup Dropdown is broken after UX update since the select control was not updated to use 'value' property directly.
...

**Checklist:**

- [ ] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/1216)


src/components/shell/deviceGroupDropdown/deviceGroupDropdown.js
```diff
@@ -9,7 +9,7 @@ import './deviceGroupDropdown.scss';
 
 export class DeviceGroupDropdown extends Component {
 
-  onChange = (deviceGroupIds) => ({ target: { value: { value } = {} } = {} }) => {
+  onChange = (deviceGroupIds) => (value) => {
```
ppathan: Is this only place where we get affected by the drop-down not getting selected correctly? Can we check if other places are fine?

isaac73: Yes I checked all other places. It's working fine.

PR #1235, opened 2019-03-15
Keyvault integration

sushilraje:
We are integrating key vault in RM. Hence, in RM deployments we would not have auth required and tenant id as environment variables. Hence, a script has been added to call keyvault and set the necessary variables before, in the EntryPoint of Docker container. The script is benign, as in, if Keyvault name, AppId, and AppSecret are available through environment variables, only then the script would execute.
# Type of change? 

- [ ] Bug fix
- [ ] New feature
- [ ] Enhancement

...

**Checklist:**

- [ ] All tests passed

---
This change is [Reviewable](https://reviewable.io/reviews/azure/pcs-remote-monitoring-webui/1235)


scripts/docker/build.cmd
```diff
@@ -33,10 +33,12 @@ cd %APP_HOME%
     xcopy /s src\*       out\docker\src\
     xcopy /s public\*    out\docker\public\
 
+    copy public\webui-config.js                     out\docker\
```
ppathan: Line 34 should be already copying webui-config.js, why do you need another copy?

scripts/docker/content/deployments/deployment-keyvault-br3-test-dotnet-basic-output.json
```diff
@@ -0,0 +1,6 @@
+{
```
ppathan: Remove this file.

scripts/docker/content/set_env.sh
```diff
@@ -0,0 +1,131 @@
...
+
+    local _temp=$(curl -X POST -H "Content-Type: application/x-www-form-urlencoded" -d "client_id=$PCS_AAD_APPID&resource=https%3A%2F%2Fvault.azure.net&client_secret=$PCS_AAD_APPSECRET&grant_type=client_credentials" $AUTH_SERVER_URL/oauth2/token)
+    AUTH_TOKEN=$(__parse_json $_temp "access_token")
+}
+
+# Fetch key vault secret.
+_get_keyvault_secret() {
+
+    # Get a new token each time you access key vault.
+    _acquire_token
```
ppathan: Can't you acquire token and save it in memory? So that you don't need to do multiple times?

scripts/docker/content/set_env.sh
```diff
@@ -0,0 +1,131 @@
...
+      # change webui config
+      modify_webui_config $_key $_value
+
+      shift
+      shift
+  done
+}
+
+main() {
+  if [[ "$PCS_KEYVAULT_NAME" != "" ]] && [[ "$PCS_AAD_APPID" != "" ]] && [[ "$PCS_AAD_APPSECRET" != "" ]]; then
```
ppathan: Are these values coming from Env vars? Can you add a comment?

scripts/docker/content/set_env.sh
```diff
@@ -0,0 +1,131 @@
...
+    sed -i "s/authEnabled.*/authEnabled: $2,/g" /app/webui-config.js
+  fi
+
+  if [ "$1" == "TENANT" ]; then
+    sed -i "s/tenant.*/tenant: '$2',/g" /app/webui-config.js
+  fi
+
+  if [ "$1" == "INSTANCE_URL" ]; then
+    if [ "$2" == "" ]; then
+      sed -i "s~instance.*~instance: 'https\:\/\/login\.microsoftonline\.com\/'~g" /app/webui-config.js
```
ppathan: Is this to insert a different login url then default?

scripts/docker/content/set_env.sh
```diff
@@ -0,0 +1,131 @@
+#!/bin/bash
+# Copyright (c) Microsoft. All rights reserved.
+
+AUTH_SERVER_URL=""
+RESOURCE_TYPE=""
+AUTH_TOKEN=""
+
+# Acquires auth token for authroizating against key vault.
+_acquire_token() {
```
ppathan: Can we use different naming conventions between method and variable names.

set_env.sh
```diff
@@ -0,0 +1,97 @@
+#!/usr/bin/env bash
```
ppathan: Why 2 copies of same file?

scripts/docker/content/set_env.sh
```diff
@@ -0,0 +1,131 @@
...
+
+#Removes "resource" prefix to extract resource type from 401 redirect of keyvault.
+__extract_resource_type() {
+    RESOURCE_TYPE=$(__extract_value_from_double_quotes $1)
+}
+
+__extract_value_from_double_quotes() {
+    local string=$1
+    # Remove trailing and prefixed double quotes.
+    local _temp=${string#*'"'}
```
ppathan: nit: better variable name, _value?

scripts/docker/content/set_env.sh
```diff
@@ -0,0 +1,131 @@
...
+    _temp=${_temp%'"'}
+
+    #return the value
+    echo $_temp
+}
+
+__parse_json() {
+  local _temp=`echo $1 | sed 's/\\\\\//\//g' | sed 's/[{}]//g' | awk -v k="text" '{n=split($0,a,","); for (i=1; i<=n; i++) print a[i]}' | sed 's/\"\:\"/\|/g' | sed 's/[\,]/ /g' | sed 's/\"//g' | grep -w $2`;
```
ppathan: nit: better variable name, _data or _value?

scripts/docker/build.cmd
sushilraje: yes not required!

scripts/docker/content/set_env.sh
sushilraje: For every request we are supposed to use new token. This is I think a security feature. That's how SDK works.

scripts/docker/content/set_env.sh
sushilraje: ok, makes sense

scripts/docker/content/set_env.sh
sushilraje: Ok.

scripts/docker/content/set_env.sh
sushilraje: yes, If we ever need to, it can be passed as an argument.

scripts/docker/content/set_env.sh
sushilraje: sure

scripts/docker/content/set_env.sh
sushilraje: This convention is used to define inner "private" functions. I just read about the conventions for bash scripts. If we have our own in this regard, I will change it to that.

scripts/docker/run.cmd
```diff
@@ -13,12 +13,20 @@ IF "%1"=="testing" goto :TESTING
...
   goto :END
 
 :TESTING
   echo Starting Remote Monitoring Web UI [testing version] ...
-  docker run -it -p 10080:10080 -p 10443:10443 %DOCKER_IMAGE%:testing
+  docker run -it -p 10080:10080 -p 10443:10443 ^
+    -e PCS_KEYVAULT_NAME ^
+    -e PCS_AAD_APPID ^
+    -e PCS_AAD_APPSECRET ^
+  %DOCKER_IMAGE%:testing
```
ppathan: nit: tab size seems different when comparing with run.sh above.

